PR #29, opened 2025-09-01
Chore/cleanup onnx tracer

Forpee:


Forpee:
> There are some warnings from unused variables but otherwise looks good to me.

Those are specific to WASM build only (see comment above)

PR #30, opened 2025-09-02
feat: Broadcasting output to match nodes input dimensions

### LINKED ISSUE #26: Make `sentiment_select` model pass e2e test
Add `JoltSNARK` workflow to the `sentiment_select` test and make it pass.
Currently fails at the Uniform Constraint level.

AntoineF4C5:
Sometimes a node's output is linked to another node's input, but their respective dimensions don't match.
While this is treated in onnx-tracer by broadcasting, it is implicit and not done in the zkvm.
This pull request aims at adding necessary `Broadcast` nodes when necessary between the first node's output and second node's input.

___
closes #26


Forpee:
Context: in `builder::multiclass0` before we added the pre-pending broadcast behavior I had to insert an explicit broadcast node
https://github.com/ICME-Lab/jolt-atlas/blob/c6f3c821eace767b46a3349f34c114ddbb9d49e1/onnx-tracer/src/builder.rs#L709-L719


when i change it to:
```rust
    // Node 5.5: Broadcast the scalar [1, 1] to [1, 10] shape
    // let scalar_broadcasted = b.broadcast(reshaped, vec![1, 10], vec![1, 10], 1);

    // Node 6: Multiply the broadcasted scalar by the weight vector (replacing RebaseScale)
    let multiplied = b.poly(PolyOp::Mult, weights_const, reshaped, vec![1, 10], 1);
```

no broadcast node is being prepended (after removing explicit `broadcast` node) in the `program_bytecode`

Can we fix this so that when we remove the explicit broadcast in `multiclass0`, the parser follows spec and inserts the broadcast

cc: @AntoineF4C5

Forpee:
> Context: in `builder::multiclass0` before we added the pre-pending broadcast behavior I had to insert an explicit broadcast node
> 
> https://github.com/ICME-Lab/jolt-atlas/blob/c6f3c821eace767b46a3349f34c114ddbb9d49e1/onnx-tracer/src/builder.rs#L709-L719
> 
> when i change it to:
> 
> ```rust
>     // Node 5.5: Broadcast the scalar [1, 1] to [1, 10] shape
>     // let scalar_broadcasted = b.broadcast(reshaped, vec![1, 10], vec![1, 10], 1);
> 
>     // Node 6: Multiply the broadcasted scalar by the weight vector (replacing RebaseScale)
>     let multiplied = b.poly(PolyOp::Mult, weights_const, reshaped, vec![1, 10], 1);
> ```
> 
> no broadcast node is being prepended (after removing explicit `broadcast` node) in the `program_bytecode`
> 
> Can we fix this so that when we remove the explicit broadcast in `multiclass0`, the parser follows spec and inserts the broadcast
> 
> cc: @AntoineF4C5

If this behavior only occurs in hand-built models (not binaries), then it’s acceptable—we’re fine with explicitly adding broadcast in those cases

PR #32, opened 2025-09-03
fix: constraint error when trace includes virtual instruction

### LINKED ISSUE #31: UniformConstraint non matching with virtual instruction
The uniform constraint `OpFlags(InlineSequenceInstruction) ⋅ (PC + NextPC - -1) == 0` does not verify when using virtual instructions IF:
- The expanded trace (where the virtual instructions are decomposed into its components) length a power of 2.

AND
- The last instruction in the trace is a virtual instruction.

The associated branch features a test that allows to reproduce this error.
Failing constraint [here](https://github.com/ICME-Lab/jolt-atlas/blob/0703a18b7f28952d11a74c88e75d493f67a42560/zkml-jolt-core/src/jolt/r1cs/constraints.rs#L318).

AntoineF4C5:
closes #31

### DIFF
--- a/onnx-tracer/src/builder.rs
+++ b/onnx-tracer/src/builder.rs
@@ -422,8 +422,10 @@ pub fn custom_addsubmuldiv_model() -> Model {
     let mut b = ModelBuilder::new(SCALE);
     let out_dims = vec![1, 4];
 
-    let x = b.input(out_dims.clone(), 2);
-    let a = b.poly(PolyOp::Add, x, x, out_dims.clone(), 2);
+    let x = b.input(out_dims.clone(), 4);
+    let a = b.poly(PolyOp::Add, x, x, out_dims.clone(), 1);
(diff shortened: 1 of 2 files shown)


Forpee:
> run `cargo test -p zkml-jolt-core addsubmuldiv` to reproduce
> 
> closes #31

Could this be fixed by pushing a no-op before adding padding (to execution trace), and then applying the pad

cc @AntoineF4C5

AntoineF4C5:
> Could this be fixed by pushing a no-op before adding padding (to execution trace), and then applying the pad

Yeah, I already have a fix available that involves that, it actually follows the logic shown here:
https://github.com/a16z/jolt/blob/87f4432c9ba75bbcc94a7fda965047a555ae269c/jolt-core/src/zkvm/mod.rs#L236-L237

Although, I still find it quite messy, but trying to make it clearer I ran into other failures which I am trying to understand.
Will still push the current working fix!

PR #36, opened 2025-09-09
Add Input in IO proof

AntoineF4C5:


zkml-jolt-core/src/jolt/tensor_heap/output_check.rs
```diff
@@ -48,29 +52,32 @@ struct OutputSumcheckProverState<F: JoltField> {
...
         debug_assert!(K.is_power_of_two());
 
-        let tensor_addresses = get_tensor_addresses(program_output.output_address);
-
-        let output_start = tensor_addresses[0];
-        let output_end = *tensor_addresses.last().unwrap();
+        // The heap layout is as follows:
+        // - [zero_register, output, input, ..memory]
+        // Hence the input-output verification range is [output_start, input_end]
+        let output_addresses = index_to_addresses(program_output.output_address);
```
AntoineF4C5, line 64: Hacky: the output is to a reserved address, and input at first address of the used memory, so input starts after output.
Will make changes so the input also goes to a reserved address so the code can be more straightforward here, it will also be more easy to port when we will allow for multiple inputs.

PR #40, opened 2025-09-17
feat: phase 3, raf, evals

Forpee:
* Adds phase 3 spartan back (pc-sum-check) which leads to invoking bytecode read-checking two more times (+ raf evals)

### DIFF
--- a/zkml-jolt-core/src/jolt/r1cs/constraints.rs
+++ b/zkml-jolt-core/src/jolt/r1cs/constraints.rs
@@ -151,44 +151,39 @@ impl<F: JoltField> R1CSConstraints<F> for JoltONNXConstraints {
             JoltONNXR1CSInputs::LookupOutput,
         );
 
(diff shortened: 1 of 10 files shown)


zkml-jolt-core/src/jolt/r1cs/constraints.rs
```diff
@@ -151,44 +151,39 @@ impl<F: JoltField> R1CSConstraints<F> for JoltONNXConstraints {
...
+        // }
+        cs.constrain_eq_conditional(
+            JoltONNXR1CSInputs::NextIsNoop,
+            JoltONNXR1CSInputs::NextUnexpandedPC,
+            0,
+        );
 
-        // // if Inline {
-        // //     assert!(NextPC == PC + 1)
+        // TODO: Rm no-op from bytecode-trace[0]
```
Forpee, line 163: Need to remove first no-op from bytecode-trace to get this to work

PR #46, opened 2025-09-30
feat: matmult (new glue logic)

Forpee:
introduces sum-check precompile proofs into the DAG system. high-level overview: since we have write-once memory we can pass val_final as advice and check it (for non-elementwise ops like matmult). we apply the standard sum-check to prove correct matmult execution. this leaves claims about a and b (the operands (we also prove c (the res) was correctly read from memory) which we prove via a read-checking sum-check. the read-checking phase shows that the values used in matmult were correctly read from memory. this leaves claims about ra_a, ra_b, ra_c (the one-hot address polys of the operands/results) and val_final. since val_final is virtual we use a further sum-check for that. also, since the operand/result addresses are hardcoded in the bytecode, we can virtualize that with a bytecode read-checking sum-check (will leave this to a follow-up PR #47)

```bash
# tests for precompile SNARK
cargo test --package zkml-jolt-core --lib -- jolt::precompiles::tests --show-output

# integration tests
cargo test -p zkml-jolt-core matmult
```


wyattbenno777 (commented):
This is hacky comments can be removed as TODO unless there is a easy way to clear it now 😄

wyattbenno777 (approved):
LGTM.

Forpee:
> LGTM.
> 
> I will probably need something like this for docs.
> 
> 1. Give C as advice (cheap!)
> 2. Prove C is correct via sum-check
> 3. Prove A, B were read correctly from memory
> 4. Prove memory is consistent
> 5. Prove addresses came from bytecode

sounds good, will do a write-up.
cc @wyattbenno777

PR #48, opened 2025-10-01
Add Sigmoid as a virtual instruction

Acentelles:
This PR introduces a new SigmoidInstruction as a virtual instruction sequence.
It implements the quantized sigmoid function:

$\sigma_Q(z)=\dfrac{Q^2}{Q+Q⋅2^{−∣z∣}}$

with Q=256.

where:
- Inputs are clamped to [−8,8] to avoid overflow in $2^z$.
- Absolute values are computed using Select(z >= 0, z, -z).
- Multiplying $Q$ by $2^{−∣z∣}$ is meant to avoid setting all $2^{−∣z∣}$ to 0 when $z > 0$.
- Branching on the sign of $z$ is handled with a Select between $Q/2^{∣z∣}$ and $Q⋅2^{∣z∣}$.
- Final computation is performed using integer-only arithmetic to avoid underflow when $z≫0$.

Key Changes
-------------

Added `SigmoidInstruction<const WORD_SIZE>` implementing VirtualInstructionSequence.

Defines a sequence of 14 virtual cycles, including:
- Clamp to $[−8,8]$ (via VirtualAdvice for now).
- Materialisation of constants (0, Q, Q²).
- Computation of abs(z) using subtraction and Select.
- Computation of 2^{|z|} via VirtualPow2.
- Integer branches: Q / 2^{|z|} and Q * 2^{|z|}.
- Sign-dependent selection into intermediate a.
- Final result: c = Q² / (Q + a) moved into td.

Run E2E tests
------------
```
cargo test --package zkml-jolt-core --lib -- jolt::e2e_tests::test_sigmoid_e2e --exact --show-output
```

### DIFF
--- a/onnx-tracer/src/ops/hybrid.rs
+++ b/onnx-tracer/src/ops/hybrid.rs
@@ -442,7 +442,7 @@ where
                 vec![LookupOp::Div {
                     denom: utils::F32((kernel_shape.0 * kernel_shape.1) as f32),
                 }]
-            },
+            }
             _ => vec![],
         }
     }
--- a/zkml-jolt-core/src/jolt/mod.rs
+++ b/zkml-jolt-core/src/jolt/mod.rs
@@ -831,7 +831,8 @@ mod e2e_tests {
     #[test]
     fn test_sigmoid_e2e() {
         // Simple case: input tensor with both positive and negative values
-        let config = ModelTestConfig::new("sigmoid", vec![3, 4, 5, 0, 0, 0, 0, 0, 0, 0], vec![1, 10]);
+        let config =
+            ModelTestConfig::new("sigmoid", vec![3, 4, 5, 0, 0, 0, 0, 0, 0, 0], vec![1, 10]);
 
         let model_fn = || model(&"../onnx-tracer/models/sigmoid/network.onnx".into());
 


Acentelles:
> NIT: Not sure if we need all of the 'u64_vec_to_i32_iter' If you already made den_td is there a reason not to use it for imm?
> 
> `imm: Some(Tensor::from(u64_vec_to_i32_iter(&den_vals))),`
> 
> https://github.com/ICME-Lab/jolt-atlas/blob/0f0b7efcc9e11a707c429fd66cb382b23238c723/zkml-jolt-core/src/jolt/instruction/sigmoid.rs#L215C1-L216C73

Thanks! No reason to reuse the already computed `u64_vec_to_i32_iter(&den_vals))`. Initially I wanted 
```
        let probs_vals: Vec<u64> = num_td
            .iter()
            .zip(&den_td)
            .map(|(&n, &d)| if d == 0 { 0 } else { n as u64 / d as u64 })
            .collect();
```
to work over `num_vals` and `den_vals` (before converting them) but division doesn't work the same way.

Acentelles:
> Does this work if zi is -2?
> 
> https://github.com/ICME-Lab/jolt-atlas/blob/0f0b7efcc9e11a707c429fd66cb382b23238c723/zkml-jolt-core/src/jolt/instruction/sigmoid.rs#L75
> 
> let z_clamped is i64 but then would turn into a big number via u64 conversion?
> 
> zi = -2 (as i64) zi as u64 = 18446744073709551614

Thanks! Good catch. I wonder if we need to precompute 1/2^a and have a `select` opcode that checks whether `a>=0`.

PR #53, opened 2025-10-06
feat: kinic + aws use-case

Forpee:
this adds two zkml examples:
1. aws — a financial transaction authorization model that decides whether to approve or deny transactions based on features like budget, trust score, amount, etc.
2. Kinic — an article classification model that returns a label based on an article snippet.

```bash
# aws
cargo run --release --example authorization

# kinic
cargo run --release --example article_classification
```

note: if this is your first time running the example, you’ll have to wait for SRS generation. subsequent runs won’t require it.


wyattbenno777:
lgtm too!

PR #56, opened 2025-10-15
First step to transformers: reducemean

Forpee:
first step toward transformers is the LayerNorm head. first opcode needed there is reduceMean, which tract does as sum + div. this adds a Sum precompile that reduces over axis=1. we’ll need a separate precompile for other axes. also added placeholders for transformer models we plan to support; reduceMean is step one

```bash
cargo test --package zkml-jolt-core --lib -- jolt::e2e_tests::test_reduce_mean --exact --show-output
```


wyattbenno777:
This does 2d but will eventually use 3d for multi-head attention?
NIT: as this is "First step"

PR #57, opened 2025-10-15
feat: broadcast

Forpee:
we can implement the broadcast node essentially as a move instruction. since dim expanding is known at preprocessing, we can calculate the read/write addresses the broadcast will touch at pp

### DIFF
--- a/onnx-tracer/src/builder.rs
+++ b/onnx-tracer/src/builder.rs
@@ -1634,3 +1634,50 @@ pub fn reduce_mean_model() -> Model {
 
     b.take(vec![input.0], vec![mean])
 }
+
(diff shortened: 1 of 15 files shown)


Acentelles:
I don't think it is implemented as an instruction in `main` as you have it here

Acentelles:
I mean that in `main`, broadcast is not implemented in jolt-zkvm-core, so it doesn't have support in the same way that this PR does, as you seemed to suggest

Forpee:
> I mean that in `main`, broadcast is not implemented in jolt-zkvm-core, so it doesn't have support in the same way that this PR does, as you seemed to suggest

not sure what you mean, since it is supported.

PR #59, opened 2025-10-16
feat: Mean of squares and select

Forpee:
Added two simple opcodes: mean_of_squares and select.
We already supported all the underlying instructions, so I just expanded mean_of_squares in onnx-tracer. The binary had it expanded already, but tract coalesces it into mean_of_squares, so I re-expanded it explicitly.
select was ported from https://github.com/ICME-Lab/zkml-jolt/pull/82

We’re now just two opcodes away from proving: [softmax, rqrt]

https://github.com/ICME-Lab/jolt-atlas/blob/8db111c2b646dc31e36727f043d4c8940ae5f57f/zkml-jolt-core/src/jolt/mod.rs#L341-L354

(test passes now because they’re currently no-ops)

to test mos and select e2e
`cargo test --package zkml-jolt-core --lib -- jolt::e2e_tests::test_self_attention_2d_transformer --exact --show-output `


AntoineF4C5 (commented):
lgtm!

PR #64, opened 2025-10-29
feat: transformer einsums

Forpee:
* Added all einsums for transformer model. `cargo test -p zkml-jolt-core --lib -- jolt::e2e_tests::test_self_attention`
* refactored read-checking to a single instance for precompiles (also reduced number of precompile polys).
* added a bunch of other tests to hammer correctness of the einsum impls

next up prove the nanogpt model. this PR sets us up for it since the transformer here is basically the same as nanogpt's


AntoineF4C5 (commented):
lgtm!

PR #79, opened 2025-11-11
refactor: trace inlining

### LINKED ISSUE #42: implement (de)serialization for `JoltSNARK`

Forpee:
* also reorganized utility functions into a dedicated utils module in onnx-tracer
* closes #42

### DIFF
--- a/zkml-jolt-core/src/utils/mcc.rs
+++ b/zkml-jolt-core/src/utils/mcc.rs
@@ -97,6 +97,15 @@ fn validate_memory_operation(
         instr,
         cycle,
     );
+    validate_read(
+        "TS3",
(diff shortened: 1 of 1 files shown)


zkml-jolt-core/src/utils/mcc.rs
```diff
@@ -0,0 +1,170 @@
...
+    validate_read(
+        "TS2",
+        cycle_index,
+        addresses.ts2,
+        cycle.ts2_read(),
+        memory,
+        instr,
+        cycle,
+    );
+
```
AntoineF4C5: Could update so it also checks value of ts3.
I can update it in the branch I'm working on and then push it too.

PR #99, opened 2025-12-11
Lookups for gelu and tanh

Forpee:
adds support for small-FP LUTs (enabled by neural teleportation)

Included in this PR
* Add GELU lookup (includes erf)
* Add tanh lookup
* Add one-hot checks

```bash
# GELU
cargo test --package zkml-jolt-core --lib -- jolt::e2e_tests::test_fnn_gelu

# Multiple GELUs in trace
cargo test --package zkml-jolt-core --lib -- jolt::e2e_tests::test_fnn_multi_gelu

# tanh
cargo test --package zkml-jolt-core --lib -- jolt::e2e_tests::test_fnn_tanh
```

Follow-up Work
A follow-up PR will add proofs that the one-hot encodings arise from the RAF polynomial.
Currently this requires a substantial feature change because:
* There is no direct 1-to-1 mapping from RAF -> RA
RAF uses signed integers (i32), and because the LUTs use <16-bit values, the sign bit changes the mapping into the field
* This means a simple i32->field->i16->field mapping cannot guarantee injectivity
* However, this mapping can be proven via a specialized sum-check

Notes
* We cannot commit to one-hot encodings until #100 is resolved.

### DIFF
--- a/zkml-jolt-core/src/jolt/mod.rs
+++ b/zkml-jolt-core/src/jolt/mod.rs
@@ -442,7 +442,7 @@ mod e2e_tests {
         let mut rng = StdRng::seed_from_u64(42);
         let scale_shift = 1 << 7; // Fixed-point scale=7 (128 = 1.0 in float)
         let input_data: Vec<i32> = (0..input_shape.iter().product())
-            .map(|_| rng.gen_range(-3..=3) * scale_shift) // Range: [-384, 384]
+            .map(|_| rng.gen_range(-3*scale_shift..=3*scale_shift) ) // Range: [-384, 384]
             .collect();
 
         // ========================================


onnx-tracer/src/graph/model.rs
```diff
@@ -1252,6 +1255,167 @@ impl Model {
...
+            };
+            new_nodes.insert(pre_idx, NodeType::Node(pre_div));
+
+            // Phase 3: Shift activation to act_idx+1, taking input from pre_idx
+            let new_act_idx = act_idx + 1;
+            if let Some(NodeType::Node(act_node)) = self.graph.nodes.get(&act_idx) {
+                let mut shifted_act = act_node.clone();
+                shifted_act.idx = new_act_idx;
+                shifted_act.inputs = vec![(pre_idx, 0)];
+                shifted_act.num_uses = 1;
```
AntoineF4C5: Keep the num_uses of original node (I don't think we use this field, but just in case).

```suggestion
```

onnx-tracer/src/graph/model.rs
```diff
@@ -1252,6 +1255,167 @@ impl Model {
...
+
+            self.graph.nodes = new_nodes;
+
+            // Patch graph outputs: shift indices and redirect activation references
+            for (out_idx, _outlet) in self.graph.outputs.iter_mut() {
+                if *out_idx > act_idx {
+                    *out_idx += 1;
+                } else if *out_idx == act_idx {
+                    *out_idx = new_act_idx;
+                }
```
AntoineF4C5:
```suggestion
                if *out_idx >= act_idx {
                    *out_idx += 1;
                }
```

zkml-jolt-core/src/jolt/fp_lookups/mod.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+/// Sumcheck instance for a single activation lookup.
+///
+/// Implements the shout protocol for small tables proving that activation outputs.
+/// Each instance handles one activation node.
+pub struct FpLookupSumcheck<F: JoltField> {
+    prover_state: Option<FpLookupProver<F>>,
+    input_claim: F,
+    num_rounds: usize,
+    tau: Vec<F>,
```
AntoineF4C5: Could we use an other name ? Such as `r_cycle`

```suggestion
    r_cycle: Vec<F>,
```

zkml-jolt-core/src/jolt/mod.rs
```diff
@@ -325,6 +348,170 @@ mod e2e_tests {
...
+        //   Downstream operations receive approximately equivalent values, preserving
+        //   model accuracy within acceptable tolerance.
+        // ============================================================================
+
+        // Generate test input: small values resembling post-layernorm outputs
+        let input_shape = [1, 30];
+        let mut rng = StdRng::seed_from_u64(42);
+        let scale_shift = 1 << 7; // Fixed-point scale=7 (128 = 1.0 in float)
+        let input_data: Vec<i32> = (0..input_shape.iter().product())
+            .map(|_| rng.gen_range(-3..=3) * scale_shift) // Range: [-384, 384]
```
AntoineF4C5:
```suggestion
            .map(|_| rng.gen_range(-3 * scale_shift..=3 * scale_shift)) // Range: [-384, 384]
```

zkml-jolt-core/src/jolt/mod.rs
```diff
@@ -325,6 +348,170 @@ mod e2e_tests {
...
+    fn test_fnn_tanh() {
+        // init_logger();
+        let input_shape = [1, 30];
+        let mut rng = StdRng::seed_from_u64(123456);
+        // Use small, zero-centered values to resemble layernorm outputs (~N(0,1)),
+        // then scale to fixed-point with out_scale=7 (quantization factor 2^7).
+        // This mimics post-layernorm quantized tensors the model expects.
+        let scale_shift = 1 << 7; // out_scale = 7
+        let input_data: Vec<i32> = (0..input_shape.iter().product())
+            .map(|_| rng.gen_range(-3..=3) * scale_shift)
```
AntoineF4C5:
```suggestion
            .map(|_| rng.gen_range(-3 * scale_shift..=3 * scale_shift))
```

PR #110, opened 2026-01-15
Feat/padding

Forpee:


atlas-onnx-tracer/src/tensor/mod.rs
```diff
@@ -406,18 +406,34 @@ impl<T: Clone + TensorType> Tensor<T> {
...
     {
         let padded_dims: Vec<usize> = self
             .dims()
             .iter()
             .map(|dim| dim.next_power_of_two())
             .collect();
 
-        let result = self.pad_to_dims(&padded_dims);
+        // Check if this is a broadcast constant (all values are the same)
+        let is_broadcast = if !self.inner.is_empty() {
```
AntoineF4C5: That's wrong, for example broadcast [10, 20], dim: [2] to [10, 20, 10, 20], dim: [2, 2]
Can you explain why you want to use this ?
Padding with pad_to_dims (i.e. with zeros) should always work

PR #117, opened 2026-01-21
feat: rsqrt

AntoineF4C5:
Implements the reciprocal square root operator.
Follows the same idea as Div operator implementation: 
We consider rsqrt as a chain of x -> 1/x -> sqrt(1/x) and prove both in the same sumcheck.
To do so we give as advice the output, intermediate output (1/x) as well as corresponding remainders.
We then prove two claims:
- Q² = x * inv + r_i
- inv * Q = rsqrt² + r_s
Where Q is the scaling factor, inv is advice value of 1/x, r_i and r_s are the remainders associated to the inverse and square root operations.

--
Note: It is left for further task to prove r_i and r_s are in the required ranges:
- 0 <= r_i < x
- 0 <= r_s < 2 * rsqrt + 1

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/ops/rsqrt.rs
+++ b/jolt-atlas-core/src/onnx_proof/ops/rsqrt.rs
@@ -18,7 +18,7 @@ use joltworks::{
         unipoly::UniPoly,
     },
     subprotocols::{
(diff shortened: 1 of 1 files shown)


Forpee (approved):
LGTM!

PR #126, opened 2026-01-23
Range Checking

### LINKED ISSUE #118: Implement range checking for `Div` and `Rsqrt`
To prove div and rsqrt, at some point we need to assert the following inequality, given two tensors a and b of equal length:
for all i, 0 <= a[i] < b[i]

The objective is to implement that, using shout and the `LT[x, y]` lookup table, which is implemented in jolt and returns 1 iff x < y.
We additionnally need to consider non-negativity of x.

### LINKED ISSUE #120: Complete non-decomp `Div` implementation
linked issue: #118

### LINKED ISSUE #121: Complete non-decomp `Rsqrt` implementation
linked issue: #118

AntoineF4C5:
This range checking implementation restricts division to using positive divisors.
We might get it to work with negative divisor without overhead through ingenious crafting for a custom Lookup Table that outputs LessThan( x, abs(y) ).

--
Closes #118, closes #120, closes #121

### DIFF
--- a/atlas-onnx-tracer/src/tensor/mod.rs
+++ b/atlas-onnx-tracer/src/tensor/mod.rs
@@ -21,6 +21,9 @@ use tract_onnx::prelude::tract_itertools::Itertools;
 /// Implementations of common operations on tensors.
 pub mod ops;
 
+// Default scale
+const SCALE: u32 = 7;
+
 /// A generic multi-dimensional array representation of a Tensor.
 /// The `inner` attribute contains a vector of values whereas `dims` corresponds to
 /// the dimensionality of the array and as such determines how we index, query for
@@ -170,9 +173,10 @@ impl Tensor<i32> {
     }
 
     /// Create a random positive tensor given the dims
+    /// The values are capped to be relatively small
     pub fn random_pos(rng: &mut StdRng, dims: &[usize]) -> Self {
         let data: Vec<i32> = (0..dims.iter().product())
-            .map(|_| (rng.next_u32() % 20000) as i32)
+            .map(|_| (rng.next_u32() % (1 << (2 * SCALE + 1))) as i32)
             .collect();
         Tensor::new(Some(&data), dims).unwrap()
     }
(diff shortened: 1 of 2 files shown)


atlas-onnx-tracer/src/tensor/mod.rs
```diff
@@ -153,23 +153,31 @@ tensor_type!(
...
             .map(|_| rng.next_u32() as i8 as i32)
             .collect();
         Tensor::new(Some(&data), dims).unwrap()
     }
 
-    /// Create a random boolean tensor give the dims
+    /// Create a random positive tensor given the dims
+    pub fn random_pos(rng: &mut StdRng, dims: &[usize]) -> Self {
+        let data: Vec<i32> = (0..dims.iter().product())
+            .map(|_| (rng.next_u32() % 20000) as i32)
```
Forpee: Where does `20,000` come from?

jolt-atlas-core/src/onnx_proof/range_checking/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+
+pub trait CommitToOneHotEncodingsMarker {
+    fn commit_to_one_encodings(&self) -> bool;
+}
+
+impl CommitToOneHotEncodingsMarker for ComputationNode {
+    fn commit_to_one_encodings(&self) -> bool {
+        matches!(self.operator, Operator::And2(_) | Operator::ReLU(_))
+    }
+}
```
Forpee: can we remove these duplicate traits and impls?

atlas-onnx-tracer/src/tensor/mod.rs
AntoineF4C5: This was taken arbitrarly, the aim was to have numbers small enough so that the value 1/x is almost never rounded to 0 when proving Recip op.
Without this restriction almost all random values taken in [0; 2^32) have a Recip of 0.
I'll set this value to Scale^2 to avoid using "magic" values.

PR #127, opened 2026-01-28
[2/2] Softmax axis=-1

### LINKED ISSUE #124: Implement `Softmax` with axis=-1

Forpee:
This PR completes the softmax verification pipeline introduced in #116  by proving the per-feature standard softmax claims required for softmax(axes = -1).

Recall that softmax over the last dimension corresponds to:
```
\prod(prev_dims) standard softmax operations
```
(i.e. one per feature vector)

This PR verifies each of those individual standard-softmax evaluations.

### Approach
For each feature vector, we verify the standard softmax computation in stages:

1. Division / Normalization
We first verify the division step of softmax as the division output yields the softmax ouput.

The sum and max checks are batched together with division:
* The sum and max tensors are scalars
* They do not need to be virtualized, since evaluating them at any point always yields the same value

This allows them to be batched with div

2. Exponentiation
Exponentiation is verified using a small LUT, following the approach used in zkGPT (except we use shout instead of lasso).

3. Operand consistency
We constrain the exponentiation raf claim to be derived from the softmax operand by verifier checking:
```
softmax_operand_claim
  = (-raf_claim) + max_logit
  ```
This also implies that the max value is greater than or equal to the other operand values, so no additional max-comparison constraint is required.
 
 4. Linking to the main operand claim
Finally, we link each per-feature softmax operand claim back to the main operand claim

---
closes #124

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/ops/softmax_axes/mod.rs
+++ b/jolt-atlas-core/src/onnx_proof/ops/softmax_axes/mod.rs
@@ -1,3 +1,35 @@
+//! Softmax verification for softmax(axes = -1).
+//!
+//! This module verifies the softmax computation over the last dimension, which corresponds to
+//! ∏(prev_dims) standard softmax operations (i.e., one per feature vector).
+//!
+//! # Verification Approach
+//!
+//! For each feature vector, we verify the standard softmax computation in stages:
+//!
+//! ## 1. Division / Normalization
+//! We first verify the division step of softmax, as the division output yields the softmax output.
+//! The sum and max checks are batched together with division:
+//! - The sum and max tensors are scalars
+//! - They do not need to be virtualized, since evaluating them at any point always yields the same value
+//!
+//! This allows them to be batched with div for efficiency.
+//!
+//! ## 2. Exponentiation
+//! Exponentiation is verified using a small LUT, following the approach used in zkGPT
+//! (except we use shout instead of lasso).
+//!
+//! ## 3. Operand Consistency
+//! We constrain the softmax operand to be derived from the exponentiation raf claim by verifier checking:
+//! ```text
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: atlas-onnx-tracer/src/tensor/ops.rs [hunk 1]
```diff
@@ -3168,17 +3168,205 @@ pub mod nonlinearities {
...
+        a: &Tensor<i32>,
+    ) -> (Tensor<i32>, Option<SoftmaxTrace>) {
+        const S: i32 = 128;
+
+        // 1) Get max
+        let max_logit = a.iter().copied().max().unwrap_or(0);
+        let max_index = a.iter().copied().position(|x| x == max_logit).unwrap_or(0);
+
+        // 2) Subtract max from each element
+        let centered_logits = a
```
AntoineF4C5: Maybe we can rename this, since rather than centering, the operation is shifting all inputs so that they are all negative, and the highest value is 0.

jolt-atlas-core/src/onnx_proof/ops/softmax_axes/softmax/exponentiation.rs
```diff
@@ -0,0 +1,685 @@
...
+    },
+};
+use rayon::prelude::*;
+use std::array;
+
+const READ_RAF_HW_DEGREE_BOUND: usize = 2;
+
+/// Shared prover/verifier parameters for Shout.
+#[derive(Clone)]
+pub struct ReadRafHwParams<F: JoltField> {
```
AntoineF4C5: It looks like this sumcheck doesn't execute HammingWeight proving, hence we can remove its mention in the name.
We can also create a HammingWeight sumcheck instance to complete proving of SoftmaxExponentiationRa.

PR #182, opened 2026-03-03
chore: rm deprecated crates

Forpee:
* rm deprecated crates `zkml-jolt-core` & `onnx-tracer`
* rename `atlas-onnx-tracer` -> `onnx-tracer`


Forpee:
To avoid disrupting current PRs, I'll postpone renaming the `atlas-onnx-tracer` crate. We can schedule this change at a more appropriate time.
in the meantime i've opened #184 which just removes the deprecated crates

Issue #138, opened 2026-02-11
If a NodeOutput poly has multiple openings, reduce openings to one


Forpee:
When traversing to a node in the prover, a NodeOutput polynomial can have multiple openings that need to be checked (virtually). We should reduce them to a single opening using the technique in PAZK 4.5.2.

PR #183, opened 2026-03-03
feat: Cos/Sin operators

### LINKED ISSUE #161: Implement `cos` op

### LINKED ISSUE #162: Implement `sin` op

AntoineF4C5:
This Pull Request implements the Cos and Sin operators.
### Core ideas
We implement those ops using a small lookup table, and previously wrapping the input around a well chosen approximation of a multiple of 2π, maintaining result accuracy thanks to the 2π-frequency of Cos/Sin functions.
To chose the constant, we considered the quantization scale of the model, currently set to 128.
This let us look for a multiple to 128·2π, that is close enough to an integer, and small enough to keep the lookup table size manageable.
This value was decided to be 128·8π = 3216.99087728 ≈ 3217.
It allows us to define a lookup table of size 4096 = 2^12, the closest power of two higher than 3217.

### Workflow
1st: Get the remainder of input division by 3217.
2nd: Lookup the remainder into the COS/SIN lookup table and output the result.

### Precision
The given workflow was tested against quantized inputs between -2^20 and 2^20 (-8192 to 8192 real values).
The average error is of 1 quantization unit ( = 0.0075 real error)
The max error is of 3 quantization units ( = 0.024 real error)

### Additional changes
These ops require a claim to be made about the remainder of step 1, and then perform a sumcheck whose input claim is taken from a linear combination of NodeOutput claim and Remainder claim.
Those claims are although not on the same opening point, so we need to evaluate one of those polynomials on the same opening point as the other.
In order to ease the path to #138, it has been chosen to compute a new opening point to NodeOutput, so that this stays the only kind of polynomial to ever require an n-to-1 opening reduction.

This caused us to exerce minor changes on the [`get_node_output_opening`](https://github.com/ICME-Lab/jolt-atlas/blob/6ceeafc6179985534214563bbc3ea4a124f30048/joltworks/src/poly/opening_proof.rs#L134) method so that it only retrieves an opening created from a node with strictly bigger index than current node. (To reserve the special-case of cos/sin where current node claims about its own output).

---

Closes #161, Closes #162

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/neural_teleport/utils.rs
+++ b/jolt-atlas-core/src/onnx_proof/neural_teleport/utils.rs
@@ -6,12 +6,13 @@
 use super::n_bits_to_usize;
 use atlas_onnx_tracer::tensor::Tensor;
 use joltworks::{
-    field::JoltField,
-    poly::eq_poly::EqPolynomial,
-    utils::thread::unsafe_allocate_zero_vec,
+    field::JoltField, poly::eq_poly::EqPolynomial, utils::thread::unsafe_allocate_zero_vec,
 };
 use rayon::{
-    iter::{IndexedParallelIterator, IntoParallelRefIterator, IntoParallelRefMutIterator, ParallelIterator},
+    iter::{
+        IndexedParallelIterator, IntoParallelRefIterator, IntoParallelRefMutIterator,
+        ParallelIterator,
+    },
     slice::ParallelSlice,
 };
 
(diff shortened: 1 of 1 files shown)


jolt-atlas-core/src/onnx_proof/ops/cos.rs
```diff
@@ -0,0 +1,669 @@
+#![allow(missing_docs)]
```
Forpee: Nit: Please remove this - we lint on missing_docs intentionally so that docs stay in sync with the code. Suppressing it per-file undermines that. Please add doc comments to the public items here instead.

Forpee (approved):
Hey, nice work getting the `Cos` and `Sin` ops landed - the overall structure is clean, the proof pipeline follows the established patterns well, and the test coverage (precision stats, boundary inputs) is solid. A few nits below that you're welcome to push back on - I don't feel strongly about any of them, they're just things I noticed.

PR #196, opened 2026-03-12
feat: Slice operator

### LINKED ISSUE #163: Implement `Slice` op

### LINKED ISSUE #197: refactor duplicate code between slice and concat
see https://github.com/ICME-Lab/jolt-atlas/pull/196#discussion_r2924206742

AntoineF4C5:
Implements `Slice` operator, supporting slicing to a unique axis, without a step params.
The reason for those choices is that tract, when encountering a Slice operator node with several axes, creates a chain of Slice nodes, each dealing with a single axis. Also if a Slice has a step value (step != 1), tract seems to rather create a Downsample node.

The workflow is very similar to `Concat` operator implemented in #187. We create a `weights` vector which holds powers of a verifier challenge ɣ. Input and output of the node are associated with a different `weights` vector, representing the mapping of input elements in the output.

### Other changes
Added a `usize` parameter to `SumcheckId::RLC` variant similar to the `NodeExecution` variant, since any node's output can be used by several nodes that require an RLC for proving.
The RLC variant however will not require a N-to-1 opening reduction, since it is not really a polynomial opening, but rather a claim to the dot-product of a polynomial with a weights tensor.

--- 
Needs #138 to constrain this node execution together with output consumer node's execution.
Closes #163, closes #197

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/ops/slice.rs
+++ b/jolt-atlas-core/src/onnx_proof/ops/slice.rs
@@ -2,35 +2,23 @@ use atlas_onnx_tracer::{
     model::trace::{LayerData, Trace},
     node::ComputationNode,
     ops::{Operator, Slice},
-    tensor::Tensor,
     utils::dims::Pad,
 };
 use common::VirtualPolynomial;
 use joltworks::{
     field::JoltField,
-    poly::{
-        multilinear_polynomial::{
-            BindingOrder, MultilinearPolynomial, PolynomialBinding, PolynomialEvaluation,
-        },
-        opening_proof::{
-            OpeningAccumulator, OpeningPoint, ProverOpeningAccumulator, SumcheckId,
-            VerifierOpeningAccumulator, BIG_ENDIAN, LITTLE_ENDIAN,
-        },
-        unipoly::UniPoly,
-    },
+    poly::opening_proof::{OpeningAccumulator, SumcheckId},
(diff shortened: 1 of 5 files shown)


jolt-atlas-core/src/onnx_proof/ops/slice.rs
Forpee: known tech debt: `GammaFold*` duplication between `slice.rs` and `concat.rs`

The `GammaFoldParams`, `GammaFoldProver`, `GammaFoldVerifier`, `gamma_powers`, and their trait impls (`SumcheckInstanceParams`, `SumcheckInstanceProver`,`SumcheckInstanceVerifier`) are duplicated across `concat.rs` and the new `slice.rs` (~290 lines). The two copies are functionally identical -- the only variation is in `initialize_input/initialize_output` which extract op-specific tensors/weights before delegating to the shared `initialize()` core

This is intentional for now to keep the slice PR self-contained and avoid refactoring concat in the same change. A follow-up should extract the shared gamma-fold machinery into `jolt-atlas-core/src/subprotocols/gamma_fold.rs`, leaving only the op specific weight ctx (`SliceGammaWeightsContext` / `ConcatGammaWeightsContext`) and the thin `initialize_input/initialize_output` wrappers in each op file.

Tracking: https://github.com/ICME-Lab/jolt-atlas/issues/197

PR #206, opened 2026-03-25
Refactor of opening handling

AntoineF4C5:
Simplify the claim caching and accessing workflow in operators proving.

In a nutshell, the proposed changes are make it more straightforward for an implementer to prove a new operator. For the claim caching, implementor now *mostly* only has to consider whether the claim is part of the graph or additionnal data given by the prover, or derived from the graph.
In the Jolt semantics, the additionnal data claims can be associated to "advices" hence the naming.

See Rsqrt example:
https://github.com/ICME-Lab/jolt-atlas/blob/8ac24139110fcfee49b7547515e24c6361bda889/jolt-atlas-core/src/onnx_proof/ops/rsqrt.rs#L340-L347
Notice the difference between a Virtual and Committed is abstracted and taken care of internally.

---

Left to be scoped by a further refactor:
- Inner workings of some operators, such as the range-checking and one-hot encoding.
- Differentiate caching behavior for Committed Polynomials, if it is OneHot or Dense (for now the helper is only used on Dense variant).


joltworks/src/subprotocols/opening_reduction.rs
```diff
@@ -362,7 +359,7 @@ impl<F: JoltField> DensePolynomialProverOpening<F> {
 
     pub fn final_sumcheck_claim(&self) -> F {
```
Forpee: to keep consistency, we should rename this as-well

jolt-atlas-core/src/utils/opening_access.rs
```diff
@@ -0,0 +1,405 @@
...
+            PolynomialId::Virtual(_) => self.acc().get_virtual_polynomial_opening(opening_id),
+            PolynomialId::Committed(_) => self.acc().get_committed_polynomial_opening(opening_id),
+        }
+    }
+}
+
+impl<'a, F: JoltField> AccOpeningAccessor<'a, F, ProverOpeningAccumulator<F>> {
+    /// Convert this accessor into a prover provider that can append openings
+    /// to the accumulator using `transcript` and `opening_point`.
+    pub fn to_provider(
```
Forpee: nit: `to_provider` consumes self, so per Rust naming [conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) this should be `into_provider`

jolt-atlas-core/src/onnx_proof/ops/einsum/mk_kn_mn.rs
```diff
@@ -249,20 +234,13 @@ impl<F: JoltField, T: Transcript> SumcheckInstanceVerifier<F, T> for MkKnMnVerif
...
         let r_right_node_output = [sumcheck_challenges.as_slice(), &r_n.r].concat();
         let right_opening_point = self.params.normalize_opening_point(&r_right_node_output);
-        accumulator.append_virtual(
-            transcript,
-            VirtualPolynomial::NodeOutput(self.params.computation_node.inputs[1]),
-            SumcheckId::NodeExecution(self.params.computation_node.idx),
-            right_opening_point,
-        );
+        provider.update_point(right_opening_point);
+        provider.append_node_io(Target::Input(1));
```
Forpee: The mutable `opening_point` on the provider is implicit state that callers must remember to update before each append at a different point. This is a footgun - forgetting `update_point` silently uses the wrong point.

consider adding `_at` variants that take the point explicitly:

```rust
provider.append_node_io(Target::Input(0), left_claim);
provider.append_node_io_at(Target::Input(1), right_claim, right_opening_point);
```

the default `append_node_io` (stored point) covers the common single-point case. The `_at` variant makes multi-point callsites self-documenting - the point is always visible next to the append it applies to

jolt-atlas-core/src/utils/opening_access.rs
```diff
@@ -0,0 +1,405 @@
...
+    where
+        Poly: Into<PolynomialId>,
+    {
+        OpeningId::new(poly_ctor(self.node.idx), self.sumcheck_id)
+    }
+}
+
+// Lightweight wrapper that lets accessors/providers work with either shared
+// or mutable accumulator references without duplicating the outer types.
+enum Reference<'a, V: ?Sized> {
```
Forpee, line 64: Reference is a runtime-checked `&`/`&mut` discriminator — `acc_mut()` panics if constructed from `&`. This works since it's module-private and only the provider path (which always gets &mut) calls `acc_mut()`. If this ever becomes harder to reason about, the alternative is splitting `AccOpeningAccessor` into separate `immutable/mutable` types so the compiler enforces it. Fine as-is for now.

joltworks/src/poly/opening_proof.rs
```diff
@@ -600,36 +605,18 @@ where
             reduced_evaluations: BTreeMap::new(),
             #[cfg(any(test, feature = "test-feature"))]
             prover_opening_accumulator: None,
-            #[cfg(feature = "zk")]
```
Forpee, line 603: rebase seems to have removed this - we should bring it back pls

joltworks/src/poly/opening_proof.rs
```diff
@@ -600,36 +605,18 @@ where
...
-            pending_claims: Vec::new(),
         }
     }
 
-    /// Takes and returns the pending ZK claims, leaving the internal vector empty.
-    #[cfg(feature = "zk")]
```
Forpee, line 609: ditto on the rebase removal. Check for other removal sites as-well pls

Forpee (approved):
LGTM!

- **Boilerplate reduction is substantial.** The repetitive `accumulator.append_virtual(transcript, VirtualPolynomial::NodeOutput(node.inputs[0]), SumcheckId::NodeExecution(node.idx), ...)` pattern is collapsed into `provider.append_node_io(Target::Input(0), claim)` - a clear win across callsites.
-  The SOC between fetching advice and node openings is well done
- **Target enum** makes intent legible at the callsite. `Target::Input(0)` vs raw index arithmetic on `node.inputs[position]`.
- **`OpeningIdBuilder`** centralizes the `SumcheckId::NodeExecution(node.idx)` default, eliminating a whole class of copy-paste bugs where the wrong node index leaks in.
- **`OpeningId` as a struct** (vs the old `enum { Committed(poly, sc), Virtual(poly, sc) }`) is better 
- **Separation of accessor (read) and provider (read + write)** is a good layering.

jolt-atlas-core/src/onnx_proof/ops/einsum/mk_kn_mn.rs
AntoineF4C5: Indeed !

PR #210, opened 2026-04-02
Fix: Add ScalarConstDivPow2 and range-check ScalarConstDiv remainders

### LINKED ISSUE #203: ScalarConstDiv remainder is committed but never range-checked
## Summary

`ScalarConstDiv` currently commits to the remainder tensor, but does not prove that the committed value is a valid remainder. This makes the fixed-point rescaling path unsound.

This issue proposes adding a sound `8-bit` range-check for the remainder while avoiding a fresh commitment for every `Mul` / `Einsum` rescaling node. Since `Mul` is one of the most frequently used operations, reducing commitments along this path is expected to substantially decrease prover time.

## Current behavior

Jolt-Atlas uses fixed-point arithmetic with an 8-bit fractional part. After multiplication, the scale is restored by inserting a `ScalarConstDiv(1 << scale)` node after `Mul` and `Einsum`.

At the moment, `ScalarConstDiv` proves only the linear identity

```text
input = divisor * output + remainder
```

and commits to the remainder tensor so that the verifier can open it at the final sumcheck point.

However, the implementation does not enforce

```text
0 <= remainder < divisor
```

As a result, the committed remainder is only checked as a polynomial opening and as part of the linear equality. It is not checked as a valid Euclidean remainder.

## Why this is a problem

Without a range-check on the remainder, an attacker can choose an invalid remainder tensor that still satisfies the linear identity. This means the proof system is incomplete for `ScalarConstDiv`, and the fixed-point rescaling path is currently unsound.

Since `ScalarConstDiv` is inserted after high-frequency operations such as `Mul` and `Einsum`, this is not an isolated corner case.

## Proposed direction

For fixed-point rescaling by `256`, the intended relation is

```text
input(i) = 256 * output(i) + rem(i)
```

As an MLE identity, this can be written as

```text
256 * output(x) = Σ_i eq(x, i) * (input(i) - rem(i))
```

so the execution sumcheck only needs the claim `rem(r)` at the final evaluation point.

To show that each entry of `rem` is 8-bit, introduce a bit-decomposition identity

```text
rem(x) = Σ_k 2^k * b(x, k)
```

If `b(x, k)` is committed before deriving the execution-sumcheck challenge, then by Schwartz-Zippel it is enough to check

```text
rem(r) = Σ_k 2^k * b(r, k)
```

at a random point `r`.

This reduces the problem to a sumcheck over the `k` direction only, so the verifier only needs the final claim `b(r, r')`.

To avoid a fresh commitment for each `ScalarConstDiv` node, replace per-node bit-decomposition polynomials with a shared polynomial

```text
B(x, t, k)
```

where `t` is a counter identifying the `ScalarConstDiv` instance.

The verifier then receives the claim `B(r, t, r')` for the `t`-th node and checks

```text
rem_t(r) = Σ_k 2^k * B(r, t, k)
```

## Expected outcome

This would:

- make `ScalarConstDiv` sound by enforcing an actual remainder range-check
- avoid per-node remainder commitments for `Mul` / `Einsum`
- reduce prover overhead for multiplication-heavy workloads

ClankPan:
This PR adds `ScalarConstDivPow2` as a dedicated path for fixed-point rebasing when the divisor is a power of two. Operations like `Mul` and `Einsum` now use this path when possible.

`ScalarConstDivPow2` represents the remainder with a one-hot RA encoding and range-checks it with the existing LUT and `commit_to_onehot()` flow. The reason for using a LUT here is that it fits the current lookup and RA machinery, so we can add this optimization without changing the commitment layer. A future `commit_to_bits()`-style path should be even lighter, but in this PR I prioritized an implementation that works with the current commitment design.

This PR also updates `ScalarConstDiv` for the general-divisor case. It now proves `0 <= r < divisor` with the same LUT/RA idea that `Div` uses. As part of that change, the dense remainder commitment is removed and replaced with a virtual remainder plus range-check claims.

One more issue showed up while testing this change. In `onnx_proof::e2e_tests::test_multihead_attention`, a `ScalarConstDivPow2` node can produce an output tensor of length 1. In that case, the RA sumcheck ends up with zero rounds and panics. Fixing this properly requires a change in the sumcheck side so that zero-round RA virtualization is handled correctly. For now, scalar `ScalarConstDivPow2` skips the RA one-hot checks, matching the current `Div` behavior. This is a real soundness issue, so it should be fixed in a separate PR.

Close #203

### DIFF
--- a/atlas-onnx-tracer/src/model/mod.rs
+++ b/atlas-onnx-tracer/src/model/mod.rs
@@ -267,7 +267,12 @@ impl Model {
                 | Operator::Rsqrt(_)
                 | Operator::Sigmoid(_)
                 | Operator::Sin(_) => LOG_K_CHUNK + log_2(node.pow2_padded_num_output_elements()),
-                Operator::ScalarConstDiv(_) => log_2(node.pow2_padded_num_output_elements()),
+                Operator::ScalarConstDiv(_) => {
+                    LOG_K_CHUNK + log_2(node.pow2_padded_num_output_elements())
+                }
+                Operator::ScalarConstDivPow2(_) => {
+                    LOG_K_CHUNK + log_2(node.pow2_padded_num_output_elements())
+                }
                 Operator::SoftmaxAxes(_) => {
                     LOG_K_CHUNK + log_2(*node.output_dims.last().unwrap_or(&1))
                 }
--- a/atlas-onnx-tracer/src/model/shadow_trace.rs
+++ b/atlas-onnx-tracer/src/model/shadow_trace.rs
@@ -442,6 +442,14 @@ fn shadow_f64(op: &Operator, inputs: Vec<&Tensor<f64>>, scale: Scale) -> Tensor<
                 elementwise_f64(inputs[0], |x| x / d)
             }
         }
+        Operator::ScalarConstDivPow2(scd) => {
+            if is_rebase_divisor(scd.divisor, scale) {
+                inputs[0].clone()
+            } else {
+                let d = scd.divisor as f64;
+                elementwise_f64(inputs[0], |x| x / d)
+            }
+        }
 
         // ── Matrix / tensor contraction ─────────────────────────────────
         Operator::Einsum(e) => tensor::ops::einsum(&e.equation, &inputs).unwrap(),
--- a/atlas-onnx-tracer/src/node/handlers/arith.rs
+++ b/atlas-onnx-tracer/src/node/handlers/arith.rs
@@ -7,7 +7,7 @@ use std::collections::HashMap;
 
 use crate::{
     node::ComputationNode,
-    ops::{Constant, Cube, Mul, Operator, ScalarConstDiv},
+    ops::{Constant, Cube, Mul, Operator, ScalarConstDiv, ScalarConstDivPow2},
     simple_handler,
     utils::{handler_builder::HandlerBuilder, parser::DecompositionBuilder},
 };
@@ -47,7 +47,7 @@ fn build_mul(hctx: &mut HandlerContext) -> Vec<ComputationNode> {
         .simple_op(Operator::Mul(Mul { scale }));
 
     #[cfg(not(feature = "fused-ops"))]
-    let builder = builder.with_auto_rebase();
+    let builder = builder.with_auto_rebase_pow2();
 
     builder.build()
 }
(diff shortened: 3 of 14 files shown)


ClankPan:
`cargo test -p jolt-atlas-core test_gpt2 -- --ignored --nocapture` was rerun locally after the latest changes and completed successfully. On my machine it finished in 446.32s total (`Proof generation took 275.18s`, `Proof verification took 143.42s`), so the CI `SIGKILL` looks like a runner timeout/cancellation issue rather than a proof failure.

Forpee:
> `cargo test -p jolt-atlas-core test_gpt2 -- --ignored --nocapture` was rerun locally after the latest changes and completed successfully. On my machine it finished in 446.32s total (`Proof generation took 275.18s`, `Proof verification took 143.42s`), so the CI `SIGKILL` looks like a runner timeout/cancellation issue rather than a proof failure.

@ClankPan Could you run the timings on your machine before these changes? The 143.42s verification time seems higher than expected, so I'd like to compare baselines

ClankPan:
> > `cargo test -p jolt-atlas-core test_gpt2 -- --ignored --nocapture` was rerun locally after the latest changes and completed successfully. On my machine it finished in 446.32s total (`Proof generation took 275.18s`, `Proof verification took 143.42s`), so the CI `SIGKILL` looks like a runner timeout/cancellation issue rather than a proof failure.
> 
> @ClankPan Could you run the timings on your machine before these changes? The 143.42s verification time seems higher than expected, so I'd like to compare baselines

@Forpee 

I ran this on my local machine and verifier looks like same time.

### main
```
Proof generation: 186.515869125s
Proof verification: 148.093474291s
Total test time: 341.06s
```

### feat/ConstScalarDiv_remainder_range_heck
```
Proof generation: 265.633045708s
Proof verification: 144.808440291s
Total test time: 433.73s
```

Forpee:
@ClankPan 
Thanks for checking. The verification times look consistent across runs, so that doesn’t appear to be the issue. On my side, I reran the gpt-2 execution profile and saw proof gen increase (gpt2 went from ~30s to ~60s), due to the reshape padding changes (#185) becoming the bottleneck (for prover and verifier) - which is interesting but out of scope for this PR, so u don't need to worry about that.

 I'll proceed with the review shortly

ClankPan:
@Forpee 
This PR #211 changes the proof method for reshape. This change will likely bring the proof time back to normal. I will rebase once it is merged.

Forpee:
Looks good. I introduced a related sub-protocol in the softmax draft PR [here](https://github.com/ICME-Lab/jolt-atlas/blob/7f8965314ccc74fb3bd17a1fa76526ffe3e18674/joltworks/src/subprotocols/identity_range_check.rs) that uses the prefix-suffix approach, which is more efficient than relying on the vanilla identity polynomial as the table. It leverages the structure in the identity table (and sparsity in ra), resulting in faster proving

Forpee:
> Fixing this properly requires a change in the sumcheck side so that zero-round RA virtualization is handled correctly.

Noted. Lets create an issue for this.

ClankPan:
> Looks good. I introduced a related sub-protocol in the softmax draft PR [here](https://github.com/ICME-Lab/jolt-atlas/blob/7f8965314ccc74fb3bd17a1fa76526ffe3e18674/joltworks/src/subprotocols/identity_range_check.rs) that uses the prefix-suffix approach, which is more efficient than relying on the vanilla identity polynomial as the table. It leverages the structure in the identity table (and sparsity in ra), resulting in faster proving

@Forpee 
Thanks for letting me know this!
I'm going to rebase this branch if the PR is merged into main.  I paused this PR until the merging.

Forpee:
resolved in #256

PR #211, opened 2026-04-02
Fix: Rewrite reshape/concat/slice with eq-selector

ClankPan:
This PR rewrites `reshape`, `concat`, and `slice` from the old gamma-fold style into an eq-selector style.

The main idea is to express reordering without adding a new commitment. At a high level, we build a selector MLE like `[eq(r, 2), eq(r, 0), eq(r, 1), ...]`, which represents an index permutation such as `0 -> 2`, `1 -> 0`, `2 -> 1`. We then use sumcheck to compute the weighted sum of the input MLE under this selector. That sum should match `out(r)`.

Because the verifier already knows the input and output shapes, it can build the same selector MLE directly and evaluate it on its own. This makes the proof flow simpler and keeps these reordering ops commitment-free.

Close #207

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/ops/concat.rs
+++ b/jolt-atlas-core/src/onnx_proof/ops/concat.rs
@@ -1,87 +1,56 @@
 use atlas_onnx_tracer::{
-    model::trace::{LayerData, Trace},
+    model::{
+        trace::{LayerData, Trace},
+        ComputationGraph,
+    },
     node::ComputationNode,
     ops::{Concat, Operator},
-    utils::dims::Pad,
 };
 use common::VirtualPolynomial;
 use joltworks::{
-    field::JoltField,
-    poly::opening_proof::{OpeningAccumulator, SumcheckId},
+    field::{IntoOpening, JoltField},
+    poly::{
+        eq_poly::EqPolynomial,
+        multilinear_polynomial::{
+            BindingOrder, MultilinearPolynomial, PolynomialBinding, PolynomialEvaluation,
+        },
+        opening_proof::{
+            OpeningAccumulator, OpeningPoint, ProverOpeningAccumulator, SumcheckId,
+            VerifierOpeningAccumulator, BIG_ENDIAN, LITTLE_ENDIAN,
+        },
+        unipoly::UniPoly,
+    },
     subprotocols::{
-        gamma_fold::{gamma_powers, GammaFoldProver, GammaFoldVerifier},
-        sumcheck::{BatchedSumcheck, SumcheckInstanceProof},
+        sumcheck::Sumcheck,
+        sumcheck::SumcheckInstanceProof,
         sumcheck_prover::SumcheckInstanceProver,
-        sumcheck_verifier::SumcheckInstanceVerifier,
+        sumcheck_verifier::{SumcheckInstanceParams, SumcheckInstanceVerifier},
     },
     transcripts::Transcript,
-    utils::errors::ProofVerifyError,
+    utils::{errors::ProofVerifyError, index_to_field_bitvector, math::Math},
 };
 
-use crate::onnx_proof::{
-    ops::{eval_reduction::NodeEvalReduction, OperatorProofTrait, ReductionFlow},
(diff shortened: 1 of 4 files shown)


jolt-atlas-core/src/onnx_proof/ops/concat.rs
```diff
@@ -94,149 +63,392 @@ impl<F: JoltField, T: Transcript> OperatorProofTrait<F, T> for Concat {
...
+                    &self.params.input_raw_dims,
+                    &self.params.output_raw_dims,
+                    self.params.axis,
+                    input_idx,
+                    &self.params.r_output.r,
+                    self.params.max_input_num_vars,
+                );
+                let selector_claim =
+                    MultilinearPolynomial::from(selector).evaluate(&full_opening_point.r);
+                let _ = input_num_vars;
```
Forpee: do we need this?

jolt-atlas-core/src/onnx_proof/ops/slice.rs
```diff
@@ -89,248 +63,349 @@ impl<F: JoltField, T: Transcript> OperatorProofTrait<F, T> for Slice {
...
-        node.idx,
-        claim_poly_input(node),
-        input_tensor,
-        ctx.input_gamma_weights(),
-        &mut prover.accumulator,
-        &mut prover.transcript,
-    )
-}
+        let input_mle = MultilinearPolynomial::from(input.padded_next_power_of_two());
+        let _output_mle: MultilinearPolynomial<F> =
```
Forpee: ditto: do we need this?

jolt-atlas-core/src/onnx_proof/ops/slice.rs
```diff
@@ -89,248 +63,349 @@ impl<F: JoltField, T: Transcript> OperatorProofTrait<F, T> for Slice {
...
+    }
 
-        if dim == self.op.axis {
-            outer_slice.resize(self.input_dims[dim] * current_slice.len(), F::zero());
-        }
+    assert!(
+        input_raw_len >= output_raw_dims.iter().product(),
+        "Slice output raw length exceeds input raw length"
+    );
+    let _ = end;
```
Forpee: ditto

jolt-atlas-core/src/onnx_proof/ops/slice.rs
```diff
@@ -89,248 +63,349 @@ impl<F: JoltField, T: Transcript> OperatorProofTrait<F, T> for Slice {
...
-fn claim_poly_output(node: &ComputationNode) -> VirtualPolynomial {
-    VirtualPolynomial::NodeOutput(node.idx)
+fn coord_to_linear(coord: &[usize], dims: &[usize]) -> usize {
+    let mut index = 0usize;
+    let mut stride = 1usize;
+    for axis in (0..dims.len()).rev() {
+        index += coord[axis] * stride;
+        stride *= dims[axis];
+    }
+    index
```
Forpee: These functions are duplicated in reshape. Please put them in shared module.

jolt-atlas-core/src/onnx_proof/ops/slice.rs
```diff
@@ -89,248 +63,349 @@ impl<F: JoltField, T: Transcript> OperatorProofTrait<F, T> for Slice {
...
-        ctx.input_gamma_weights(),
-        &mut verifier.accumulator,
-        &mut verifier.transcript,
-    )
-}
+    fn compute_message(&mut self, _round: usize, previous_claim: F) -> UniPoly<F> {
+        const DEGREE_BOUND: usize = 2;
+        let half_poly_len = self.input_mle.len() / 2;
+        let mut uni_poly_evals = [F::zero(); 2];
+        for i in 0..half_poly_len {
```
Forpee: parallelize this please (also in reshape and concat). See other ops sum-check's for implementation details.

jolt-atlas-core/src/onnx_proof/ops/concat.rs
ClankPan: Thanks! I fixed them

PR #212, opened 2026-04-02
Feat: gather for small and large

ClankPan:
This PR adds a split between `GatherSmall` and `GatherLarge`.

Qwen uses a vocabulary size that is too large for the existing `Gather` proof path, because the one-hot address representation assumes the table size fits in `u16`. To handle this, `Gather` is now split by dictionary size. Small tables continue to use the existing path, while large tables are lowered to `GatherLarge`, which uses chunked one-hot encoding.

The verifier can make the same choice because it already knows the node shape and dictionary size from preprocessing. So both prover and verifier can consistently choose `GatherSmall` or `GatherLarge` from the graph.

With this change, Qwen no longer fails at the old gather table-size limit. Qwen still does not run end-to-end yet, because the proof system is still missing some required `Einsum` patterns.


jolt-atlas-core/src/onnx_proof/ops/einsum/mod.rs
```diff
@@ -45,6 +45,19 @@ pub mod mbk_nbk_bmn;
 /// Einstein summation for matrix-matrix multiply: mk,kn->mn
 pub mod mk_kn_mn;
 
+// TODO(Qwen): add proof implementations for the remaining einsum patterns that
```
Forpee: nit: rm

joltworks/src/subprotocols/shout.rs
```diff
@@ -34,6 +34,9 @@ fn resolve_vp_opening<F: JoltField>(
     vp: VirtualPolynomial,
     sid: SumcheckId,
 ) -> (OpeningPoint<BIG_ENDIAN, F>, F) {
+    if let Some(opening) = accumulator.try_get_virtual_polynomial_opening(vp, sid) {
```
Forpee, line 305: nice fallback

jolt-atlas-core/src/onnx_proof/witness.rs
Forpee, line 267: nit:  like `GatherRaD` does below, we should read `num_words` from `gather_op.dict_len`for consistency

atlas-onnx-tracer/src/model/shadow_trace.rs
```diff
@@ -479,7 +479,11 @@ fn shadow_f64(op: &Operator, inputs: Vec<&Tensor<f64>>, scale: Scale) -> Tensor<
...
         // ── Index / Lookup ──────────────────────────────────────────────
-        Operator::Gather(g) => {
+        Operator::GatherSmall(g) => {
+            let indices: Tensor<usize> = inputs[1].map(|v| v as usize);
+            tensor::ops::gather(inputs[0], &indices, g.axis).unwrap()
+        }
+        Operator::GatherLarge(g) => {
             let indices: Tensor<usize> = inputs[1].map(|v| v as usize);
             tensor::ops::gather(inputs[0], &indices, g.axis).unwrap()
         }
```
Forpee, line 489: nit: these two arms can be collapsed into one

Cargo.toml
```diff
@@ -37,6 +37,7 @@ num-derive = "0.4.2"
 num-traits = "0.2.19"
 
 # Serialization
+bincode = { version = "2.0.1", features = ["serde"] }
```
Forpee, line 40: nit: if this is only used in the Qwen example we should put it under `[dev-dependencies]` so it doesn't ship with the library

Forpee (approved):
LGTM - the split is clean!

AntoineF4C5:
@ClankPan I'm curious why do we need to keep the old Gather behavior as well ? Is the overhead of using chunked one-hot encodings big for small tables ? 
I believe for simplicity, if the overhead is negligible, we can stick to a unique implementation that uses chunked encodings.
Curious to hear your thoughts @Forpee.

ClankPan:
@AntoineF4C5 
Thank you for the review.

The `gather` instruction can have a table size of 1 in cases such as BGE. As a result, the number of sumcheck rounds becomes 0, which causes the current implementation to panic. For that reason, I kept `GatherSmall` so that the existing tests would still pass.

Forpee:
Thanks @AntoineF4C5  - I agree that converging on a single gather implementation using decomposed ra would be preferable for simplicity

As Clank pointed out, we currently hit an edge case in ra virtualization when the table size is 1 (and also 2), which causes failures

i actually suggested earlier that we track this explicitly in a follow-up issue (https://github.com/ICME-Lab/jolt-atlas/pull/210#issuecomment-4178871942). I'd propose we proceed with the current workaround for now and track fixing this limitation in a follow-up issue

AntoineF4C5:
Makes sense ! Thank you for the context.

Let's add an issue then. Other than that, the PR looks good to me!

AntoineF4C5 (approved):
LGTM

Issue #192, opened 2026-03-09
Prove tanh via small LUT + (out-of-range -> ±scale)


Forpee:
Generate a tanh lookup table for an arbitrary fixed-point scale S.

Entry i = round(tanh(i / S) * S) for i ∈ [0, table_size).

Only the non-negative half of the table is stored. Since tanh is an odd function, the caller negates the result for negative inputs.

tanh(x) saturates to ±1 for |x| ≳ 4, so the table only needs to cover [0, 4·S]. For inputs beyond this range, the output is clamped to ±S.

Issue #61, opened 2025-10-23
Implement clamping instruction


AntoineF4C5:
A virtual instruction that clamp inputs to be between two given values.
Should not be too hard if the limits are powers of two.

PR #215, opened 2026-04-07
feat: softmax opt + rqe changes

### LINKED ISSUE #175: Reduce proving/verification overhead of Softmax
### Context
To perform softmax in the virtual machine, we first get the maximum value of each vector of the axis on which softmax is performed.
We then offset all values by this maximum value, to ensure all have the same sign.
We then compute the exponential e^(max-x) through a lookup to a `EXP` table, compute the sum of all e^x for each vector, to finally compute the softmax.

### Current behavior
Currently for softmax against last axis of a [head, sequence, feature] tensor, we perform most of the above computation separately for each of the `head * sequence` features and build each [feature] output shard, and then use all the outputs from those parallel computations to rebuild the complete [head, sequence, feature] output.
This means that we perform a lookup to the `EXP` table for each of the `head * sequence` feature tensors.
However each lookup requires defining a one-hot encoding of a `feature` elements vector, and the one-hot encoding is a `word_size * feature` elements, that needs to be constrained using shout.
All sumchecks needed to prove the lookup of each `feature` vector are batched, but this might still incure a big overhead compared to performing all computation once on the whole [head, sequence, feature] tensor.

### Proposed proving flow
Perform all the computation mentionned above without decomposing the tensor into [feature] vectors, hence for example to recover the max of each feature, we will return a [head, sequence, 1] tensor that holds the max for each. We can then recover the offset values from this max tensor, and feed the whole offset_tensor to the `EXP` lookup table, which will in turn require to prove a `word_size * head * sequence * feature` one-hot encoding.
We follow the same idea to sum all elements of each feature vectors and recover the complete output directly, since we never had to decompose the tensor into [feature] shards.

### LINKED ISSUE #191: Prove softmax exp via small exp LUT + (out-of-range -> 0)
The range of indices in the exp LUT that produce non-zero outputs is very small. Therefore, we can use a small LUT but enforce that any index outside the smaller exp LUT range returns 0

Forpee:
Updates softmax to follow Antoine's suggestion: run a single sumcheck over the F×N variables instead of F sumchecks over N variables.

spec: https://hackmd.io/@R5TO3fi7TlWQKnY6Ejr2eg/Byjucoqjbx

## Softmax Proving: 27× Speedup

Replaced `SoftmaxAxes::prove` with `SoftmaxLastAxis::prove`. Benchmarked on GPT-2 (12 softmax layers):

|  | Old (`SoftmaxAxes`) | New (`SoftmaxLastAxis`) | Speedup |
|---|---|---|---|
| Per-layer avg | ~1,088 ms | ~41 ms | **27×** |
| Total (12 layers) | 13.05 s | 0.49 s | **27×** |

Each softmax prove dropped from **~1.1 s -> ~41 ms**, saving **~12.6 s** total on GPT-2.

---

Closes #191
Closes #175  

Follow-up work:
*  #216
* Bumped tanh `LOG_TABLE_SIZE` from 12 -> 16. The new softmax path produces slightly larger lookup indices, so the previous table size is insufficient. 
Follow-up: #192 will address this and restore efficiency.


Forpee:
After softmax changes the current op with longest proof gen in gpt2: tanh

atlas-onnx-tracer/src/model/mod.rs
```diff
@@ -268,8 +270,8 @@ impl Model {
                 | Operator::Sigmoid(_)
                 | Operator::Sin(_) => LOG_K_CHUNK + log_2(node.pow2_padded_num_output_elements()),
                 Operator::ScalarConstDiv(_) => log_2(node.pow2_padded_num_output_elements()),
-                Operator::SoftmaxAxes(_) => {
-                    LOG_K_CHUNK + log_2(*node.output_dims.last().unwrap_or(&1))
+                Operator::SoftmaxLastAxis(_) => {
+                    LOG_K_CHUNK + log_2(node.pow2_padded_num_output_elements())
```
AntoineF4C5: We can group this variant's behavior with Sigmoid, Sin ... variants.

atlas-onnx-tracer/src/model/mod.rs
```diff
@@ -463,12 +465,13 @@ impl RunArgs {
...
-    pub const DEFAULT_SCALE: i32 = 7;
+    pub const DEFAULT_SCALE: i32 = 8;
 
-    /// Const used as an approximation of 8π in the zkVM with a scale factor of 128.
+    /// Const used as an approximation of 8π in the zkVM with a scale factor of 2^DEFAULT_SCALE.
     /// This allows us to wrap around a multiple of 2π, using the periodicity of the cosine function
     /// while keeping the lookup table size reasonable.
-    pub const EIGHT_PI_APPROX: i32 = 3217;
+    /// Value: round(8 * π * 2^DEFAULT_SCALE) = round(8 * π * 256) = 6434
+    pub const EIGHT_PI_APPROX: i32 = 6434;
```
AntoineF4C5: The previous multiple chosen for pi was 8, because it was the smallest multiple close enough to an integer (consider 128*pi as being 1pi, due to quantization). Hence the number was computed as 8 * pi * 128 ≈ 3216.99, which is the closest we can get to an integer with small multiples.
Fortunately, the multiple 8 * pi * 128 equals the multiple 4 * pi * 256.
Which means we can keep the same approximation, and table size, however this multiple is now 4pi, and not 8pi.


```suggestion
    /// Const used in the zkVM to wrap around a multiple of 2π, 
    /// using the periodicity of the cosine function while keeping the lookup table size reasonable.
    /// For a scale of 8, the best multiple was found to be 4: value = round(4 * π * 256) = 3217
    pub const FOUR_PI_APPROX: i32 = 3217;
```

atlas-onnx-tracer/src/model/test.rs
```diff
@@ -222,7 +230,7 @@ impl ModelBuilder {
             id,
             Operator::Einsum(Einsum {
                 equation: equation.to_string(),
-                scale: SCALE as i32, // TODO: Pass in scale from runtime args instead of hardcoding here.
+                scale: DEFAULT_SCALE as i32, // TODO: Pass in scale from runtime args instead of hardcoding here.
```
AntoineF4C5, line 233: Can you explain why for this op you cannot use `1 << self.scale` as in other operators (such as `softmax_last_axis`) ?

atlas-onnx-tracer/src/ops/softmax.rs
```diff
@@ -1,47 +1,295 @@
...
+    pub last_dim: usize,
+    /// X[k,j] — input logits, flat [F*N]
+    pub x: Vec<i32>,
+    /// max_k — max of feature vector k, [F]
+    pub max_k: Vec<i32>,
+    /// argmax_k[k] — position of the max in each feature vector, [F]
+    pub argmax_k: Vec<usize>,
+    /// e_k[j] — one-hot indicator at argmax_k, flat [F*N] (1 at argmax, 0 elsewhere)
+    pub e_k: Vec<i32>,
+    /// z[k,j] — max_k − X[k,j] (≥ 0), flat [F*N]
```
AntoineF4C5: I wonder if you rather meant:
A few of those fields comments are confusing, interverting the "-" as a dash and as a subtraction operator.

```suggestion
    /// z[k,j] = max_k − X[k,j] (≥ 0), flat [F*N]
```

atlas-onnx-tracer/src/ops/softmax.rs
```diff
@@ -1,47 +1,295 @@
...
+pub fn softmax_last_axis_decomposed(
+    a: &Tensor<i32>,
+    scale: i32,
+) -> (Tensor<i32>, SoftmaxLastAxisTrace) {
+    let dims = a.dims();
+    let last_dim = *dims.last().unwrap();
+    let num_slices: usize = dims.iter().product::<usize>() / last_dim;
+    let data = a.data();
+    debug_assert!(
+        scale.trailing_zeros() <= 15,
```
AntoineF4C5: Seems like a confusing way to assert that scale <= 2^15, can you explain that choice?

atlas-onnx-tracer/src/ops/softmax.rs
```diff
@@ -1,47 +1,295 @@
...
+            let z_lo = (zu & (decomp.base - 1)) as i32;
+            w_z_hi[idx] = z_hi;
+            w_z_lo[idx] = z_lo;
+
+            // Sub-table lookups
+            let hi_val = if (z_hi as usize) < decomp.lut_hi.len() {
+                decomp.lut_hi[z_hi as usize]
+            } else {
+                0
+            };
```
AntoineF4C5: This test should always return true, due to definition of z_c as `min(z[idx], z_bound-1)`

```suggestion
            let hi_val = decomp.lut_hi[z_hi as usize];
```

atlas-onnx-tracer/src/ops/softmax.rs
```diff
@@ -1,47 +1,295 @@
...
+            } else {
+                let product = hi_val as i64 * lo_val as i64;
+                exp_q[idx] = (product / s as i64) as i32;
+                w_r_exp[idx] = (product - exp_q[idx] as i64 * s as i64) as i32;
+                debug_assert!(
+                    w_r_exp[idx] >= 0 && w_r_exp[idx] < s,
+                    "r_exp out of range: {}, S={s}",
+                    w_r_exp[idx]
+                );
+            }
```
AntoineF4C5: Unnecessary case disjunction: 
If hi_val == 0, the `else` case would also set both exp_q and w_r_exp to 0.

```suggestion
            let product = hi_val as i64 * lo_val as i64;
            exp_q[idx] = (product / s as i64) as i32;
            w_r_exp[idx] = (product - exp_q[idx] as i64 * s as i64) as i32;
            debug_assert!(
                w_r_exp[idx] >= 0 && w_r_exp[idx] < s,
                "r_exp out of range: {}, S={s}",
                w_r_exp[idx]
            );
```

common/src/lib.rs
```diff
@@ -1,579 +1,355 @@
...
 pub mod utils;
 
-#[derive(Hash, PartialEq, Eq, Copy, Clone, Debug, PartialOrd, Ord, Allocative)]
-pub enum CommittedPolynomial {
-    /// Fields:
+// ---------------------------------------------------------------------------
+// CommittedPolynomial
+// ---------------------------------------------------------------------------
+
+canonical_serde_enum! {
```
AntoineF4C5, line 10: Thank you for the helper!

jolt-atlas-core/src/onnx_proof/ops/softmax_last_axis/mod.rs
```diff
@@ -0,0 +1,1344 @@
...
+    fn send_auxiliary_vectors<F: JoltField, T: Transcript>(&self, prover: &mut Prover<F, T>) {
+        let [f, _] = self.F_N;
+        for k in 0..f {
+            prover.accumulator.append_virtual(
+                &mut prover.transcript,
+                VirtualPolynomial::SoftmaxSumOutput(self.node_idx, k),
+                SumcheckId::NodeExecution(self.node_idx),
+                OpeningPoint::default(),
+                F::from_u32(self.trace.exp_sum_q[k] as u32),
+            );
```
AntoineF4C5, line 289: I do not really like the idea of caching all polynomial evaluations 1 by 1 to send the polynomial to the verifier.
I would prefer to explicitely send the vector/polynomial to the accumulator by adding a new field dedicated to that.

jolt-atlas-core/src/onnx_proof/witness.rs
```diff
@@ -395,6 +333,96 @@ impl<F: JoltField> WitnessGenerator<F> for CommittedPolynomial {
...
+    };
+    let layer_data = Trace::layer_data(trace, node);
+    let input = &layer_data.operands[0];
+    softmax_last_axis_decomposed(input, *scale).1
+}
+
+/// Build a one-hot RaD witness for a SoftmaxLastAxis identity range-check or
+/// Shout ra polynomial, given raw lookup indices and the log2 of the table /
+/// range-check size.
+fn build_softmax_last_axis_onehot_witness<F: JoltField>(
```
AntoineF4C5: This method seems to not be exclusive to softmax-related one-hot encodings, we can maybe rename to a more generic name (and reuse for other one-hot encodings ?).

AntoineF4C5 (changes requested):
Great implementation ! There is a lot of nice tricks that you are using here.

I'm unsure about sending whole tensors for max_k, exp_sum_q and argmax_k to the verifier, but if we really want to push forward we can leave that for now, and address the issue later.

I also want to give another pass at the PR to ensure more thoroughly that everything is well enforced.

atlas-onnx-tracer/src/ops/softmax.rs
Forpee: Thanks - dead code from an earlier iteration, removed.

atlas-onnx-tracer/src/ops/softmax.rs
Forpee: Good point - replaced with `scale <= (1 << 15)`.

atlas-onnx-tracer/src/model/test.rs
Forpee: Artifact from a previous impl. The scale representation is a mess across ops right now - three different conventions:

* `i32` log-scale: Mul, Einsum, Cube, MeanOfSquares
`F32` (multiplier 2^scale): Cos, Sin, Tanh, Sigmoid, Erf
* `i32` multiplier: SoftmaxLastAxis
* Rsqrt is its own thing (stores log-scale as `F32`)

Will unify in a follow-up.

atlas-onnx-tracer/src/model/mod.rs
Forpee: Good catch - renamed to `FOUR_PI_APPROX` with value 3217 (4*pi * 256).

jolt-atlas-core/src/onnx_proof/ops/softmax_last_axis/mod.rs
Forpee: Deferring - since we plan #218 to remove the auxiliary vectors entirely

Forpee:
@AntoineF4C5 Thanks for the thorough review! All the nits have been addressed - FOUR_PI_APPROX changes, dead code removal, doc fixes, build_onehot_witness rename, etc.

Re the auxiliary vectors: agreed, they're not ideal. Added a module-level doc + TODO(#218) to track removing them

ClankPan:
> I'm unsure about sending whole tensors for max_k, exp_sum_q and argmax_k to the verifier, but if we really want to push forward we can leave that for now, and address the issue later.

@AntoineF4C5 
I was also concerned about this, so I looked into the proof-size increase.
It is given by `head_len * seq_len * num_softmax * 3`, so for Qwen this becomes `1008 * seq_len * 32 bytes`.
That means for 64 tokens, it comes out to about `2 MiB`. It's a little big, but I think it's within an acceptable range

atlas-onnx-tracer/src/model/test.rs
AntoineF4C5: You're right, there are indeed quite a few different ways used to represent the same thing, it'll be good to unify.

atlas-onnx-tracer/src/ops/softmax.rs
```diff
@@ -1,47 +1,286 @@
...
+    pub max_k: Vec<i32>,
+    /// argmax_k[k] — position of the max in each feature vector, [F]
+    pub argmax_k: Vec<usize>,
+    /// e_k[j] — one-hot indicator at argmax_k, flat [F*N] (1 at argmax, 0 elsewhere)
+    pub e_k: Vec<i32>,
+    /// z[k,j] = max_k − X[k,j] (≥ 0), flat [F*N]
+    pub z: Vec<i32>,
+    /// exp_q[k,j] — LUT[z[k,j]], flat [F*N]
+    pub exp_q: Vec<i32>,
+    /// exp_sum_q[k] — Σ_j exp_q[k,j], [F]
```
AntoineF4C5: There is a similar situation in the fields comment here, where the dash can be mistaken as a subtraction operator.
Can we also replace that, most likely by an "equal" operator ? 
Same will need to be done for most, if not all, of the other fields of this struct.

```suggestion
    /// exp_sum_q[k] = Σ_j exp_q[k,j], [F]
```

atlas-onnx-tracer/src/ops/softmax.rs
```diff
@@ -1,47 +1,286 @@
...
+    pub z_hi: Vec<i32>,
+    /// z_lo[k,j] = z[k,j] & (B-1) — low digit of centered logit
+    pub z_lo: Vec<i32>,
+    /// exp_hi[k,j] = LUT_hi[z_hi[k,j]]
+    pub exp_hi: Vec<i32>,
+    /// exp_lo[k,j] = LUT_lo[z_lo[k,j]]
+    pub exp_lo: Vec<i32>,
+    /// r_exp[k,j] = exp_hi·exp_lo − exp_q·S  ∈ [0, S)
+    pub r_exp: Vec<i32>,
+    /// sat_diff[k,j] = z[k,j] − z_c[k,j]  (≥ 0, saturation overflow)
```
AntoineF4C5: z_c is never mentioned in the trace fields or doc, I believe it is equal to z, clamped in i32 boundaries, but might be worth mentioning.

joltworks/src/poly/opening_proof.rs
```diff
@@ -933,6 +933,8 @@ pub enum SumcheckId {
     /// A node-execution sumcheck for a specific consumer node.
     /// The payload is the consumer's node index.
     NodeExecution(usize),
+    /// Arbitrary sum-check id for sum-checks not associated with a specific node execution, e.g. for the different stages in softmax proof.
+    Execution,
```
AntoineF4C5: I believe the sumchecks executed to prove softmax are still related to the Node that performs softmax, so we can maybe use NodeExecution sumcheck_id ?
That would avoid creating a new variant.
Also could disambiguate, since both have a similar name

jolt-atlas-core/src/onnx_proof/ops/softmax_last_axis/mod.rs
```diff
@@ -0,0 +1,1350 @@
...
+        &self,
+        node: &ComputationNode,
+        verifier: &mut Verifier<'_, F, T>,
+    ) -> Result<(), ProofVerifyError> {
+        SoftmaxLastAxisVerifier::new(node, self.scale, verifier).verify(node, verifier)
+    }
+
+    fn get_committed_polynomials(&self, node: &ComputationNode) -> Vec<CommittedPolynomial> {
+        // self.scale is the actual scale S (e.g. 4096); log_scale is log₂(S) (e.g. 12)
+        let log_scale = self.scale.trailing_zeros() as usize;
```
AntoineF4C5: Is there a specific reason why not to use ilog2, which also matches better the way log_hi and log_lo are derived ?

```suggestion
        let log_scale = self.scale.ilog2() as usize;
```

jolt-atlas-core/src/onnx_proof/ops/softmax_last_axis/mod.rs
```diff
@@ -0,0 +1,1350 @@
...
+            &prover.accumulator,
+            &mut prover.transcript,
+        );
+        // Clone exp_q for exp_sum, take for recip_mult (two consumers need owned copies)
+        let exp_q_for_sum = self.trace.exp_q.clone();
+        let recip_mult_prover = RecipMultProver::initialize(
+            std::mem::take(&mut self.trace.exp_q),
+            std::mem::take(&mut self.trace.inv_sum),
+            recip_mult_params,
+        );
```
AntoineF4C5: inv_sum is sent both to the "params" and "prover" of RecipMult, we can restrict to sending only to one of them (maybe just send it to the prover, and add an "inv_sum_aux" field to the verifier ?)

jolt-atlas-core/src/onnx_proof/ops/softmax_last_axis/exponentiation/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /// vp type for raf polynomials (`z_hi` or `z_lo`).
+    fn raf_vp(self, node_idx: usize) -> VirtualPolynomial {
+        match self {
+            Self::Hi => VirtualPolynomial::SoftmaxExpZHi(node_idx),
+            Self::Lo => VirtualPolynomial::SoftmaxExpZLo(node_idx),
+        }
+    }
```
AntoineF4C5, line 42: I suggest renaming those to remove the "exp" of the name, since they refer to z, and can be mistaken for the polynomial refering to exp^z, built later.

```suggestion
    fn raf_vp(self, node_idx: usize) -> VirtualPolynomial {
        match self {
            Self::Hi => VirtualPolynomial::SoftmaxZHi(node_idx),
            Self::Lo => VirtualPolynomial::SoftmaxZLo(node_idx),
        }
    }
```

AntoineF4C5 (commented):
I added a few nits concerning the implementation.

I also have 2 soundness concerns for this proof:
### Prover crafting an auxiliary to trick the verifier
In the proof flow, the first thing the prover does is send the auxiliaries to the verifier, by using the accumulator caching - which ensures the transcript is updated with it, and the prover can not later modify the auxiliaries to adapt to verifier challenges created during the softmax proving.

However, the auxiliary caching still happens after "r0" (i.e the opening point on softmax's output) has been sent by the verifier.
I have not digged more as to whether that be exploited by a malicious prover, but this would need to be asserted.

Namely, concerning the auxiliary `exp_sum`, it is used in two instances: 
Note `inv_sum = 1/ exp_sum` is also built by the verifier given exp_sum.
```
softmax_x(r0) = ∑_i eq(i, r0) * exp_z(i) * inv_sum(i)
exp_sum(r0) = ∑_i eq(i_k, r0_k) * exp_z(i)
```
Since inv_sum and exp_sum are MLE of the original polynomials, the property that inv_sum(r) = 1 / exp_sum(r) does not hold outside of the boolean hypercube.
It is possible that a malicious prover could exploit this to craft a "exp_sum" tensor to pretend any input maps to any output of a softmax node.

### z unlinked to z_hi, z_lo
I did not really see a link between z and z_hi, z_lo.
I was looking for something like `z - sat_diff = z_hi * S + z_lo`.
The range check that z_lo < S should suffice for uniqueness of the above writing.

Another note, I am not sure the "sat_diff" is sufficiently constrained. In fact I think a malicious prover could use any positive value he wants, as long as `z_c < z_bound` (it shouldn't be possible to have `sat_diff != 0` if `z_c != z_bound`)

Forpee:
@AntoineF4C5 

Thanks for triple checking the soundness of the protocol here are my thoughts:

the prover see's `r0` before sending `exp_sum_q` - that's true. but the claim `softmax_q(r0)` is already fixed: some upstream node consumed the softmax output as an operand, ran its own sumcheck, and reduced to a claim `softmax_q(r0)` at a random point. That claim is provided as the input to Stage 1's recip-mult sumcheck. A fake `exp_sum_q -> fake inv_sum -> different softmax_q `values on the hypercube -> different mle poly. i.e. : if the prover uses fake intermediates, the resulting softmax values on the hypercube define a different polynomial, which disagrees with the pinned evaluation at `r0`.  so by sz this matches the upstream-pinned `softmax_q(r0)` with probability `log_2(F * N) / |Field|` - negligible the prover can't "adapt" the auxiliary to a claim that was already derived from an upstream verification step.
That said we can change it to send the aux polys before challenge `r0` is sent.


* On "z unlinked to z_hi, z_lo":

The link is the operand link - you may have missed it. The verifier checks:

```
X(r2) = max_k(r2_k) − z_hi(r2) * B − z_lo(r2) − sat_diff(r2)
```

rearrage: 
```
max_k(r2_k) − X(r2) = z_hi(r2) * B + z_lo(r2) + sat_diff(r2)
```
the LHS is `z(r2)`  (i.e. `max_k - X`)

which is exactly `z = z_hi * B + z_lo + sat_diff` enforced at a random point via sz. There is no separate `z` or `z_c` polynomial - `z` is implicitly `max_k − X` (the LHS), and `z_c = z_hi * B + z_lo` is implicitly the RHS minus `sat_diff`. The RAF checks force `z_hi < K_hi` and `z_lo < B`, which gives uniqueness of the  decomposition for any given `z_c`


* on "`sat_diff` insufficiently constrained":

Changing `sat_diff` while maintaining the operand link forces a different `z_c = z_hi *B + z_lo`, which changes the  lookups (`exp_hi = LUT[z_hi], exp_lo = LUT[z_lo]`), which changes `exp_q`, which changes `softmax_q`. But `softmax_q(r0)` is already pinned by the upstream consumer's sumcheck -> the prover can't retroactively change it. soo `sat_diff` is transitively constrained through the full pipeline

jolt-atlas-core/src/onnx_proof/ops/softmax_last_axis/mod.rs
Forpee: good catch - done!

joltworks/src/poly/opening_proof.rs
Forpee: Agreed. removed it.

AntoineF4C5:
@Forpee My pleasure to align our understanding.

Concerning point 1:
> that's true. but the claim softmax_q(r0) is already fixed
This is in fact where I believe the issue is. Since softmax_q(r0) is already fixed, the prover can manipulate exp_sum_q to modify it to it's liking, while preserving the dot-product equality.

Concerning point 2:
Indeed, I had missed that, thank you for pointing it! I think I missed it because it was not in any of stages 1-4 flows.

Concerning point 3:
The idea, similar to point 1, is not to change the output, but rather to start from a valid output, and pretend this output comes from a wrong input.
In this case, we know the valid `z_c` which allows to continue pathing towards our valid output. This means the fixed variable is z_c, and when unsufficiently constrained, anything else can be played with by the prover to trick the verifier into an accepting a wrong input.

Here, the constrains on `z_c` are: `z_c = max_x - x - sat_diff`, with `sat_diff > 0` and `x < max_x`. This means the prover can adapt the X input to softmax, with a matching `sat_diff`, so that the verifier believes it corresponds to the valid output.

Forpee:
@AntoineF4C5 

on point 1 (working backwards from pinned output):

Right that the prover can try work backwards from the pinned `softmax_q(r0)`, but the constraint chain is pinned on both ends - not just the output.

`X(r2)` is produced by the max-indicator sum-check as a `NodeOutput(softmax.inputs[0]`) claim. this flows upstream through virtualization chain: it becomes the input claim for an upstream sum-check, which reduces further upstream, and so on, until it reaches the model input that the verifier evaluates directly. So a malicious prover can't pick an `X'` thats inconsistent with the actual input - it would disagree with an evaluation somewhere along the chain.

The protocol is pinned on both ends:
* Output: `sofmax_q(r0)` pinned by the downstream consumers sum-check
* Input: `X(r2)` pinned by the upstream virtualization claim

i.e. `X(r2)` is a `NodeOutput` claim that the verification chain must reconcile with the verifier-known model input. The prover can't freely choose it because it's constrained by the full chain, not just by the softmax protocol in isolation.

with both ends fixed, the intermediates (`exp_sum_q`, `inv_sum`, `exp_q`, etc...) are uniquely determined by the sum-check relations + range-checks. theres not room to manipulate the auxilary.

That said, we can move the aux sends before `r0` if you'd prefer - it's a simple change and removes the ordering concern entirely.

* On point 3 (adapt `X` and `sat_diff` while preserving `z_c`):

This is the same attack as point 1, made concrete. The operand link is:

```
max_k[k] − X[k,j] = z_hi[k,j]·B + z_lo[k,j] + sat_diff[k,j]
```

your attack: keep `z_c = z_hi * B + z_lo` fixed (valid output preserved), pick `X' = X - delta` and `sat_diff' = sat_diff + delta`. Arithmetically the operand link still holds - thats correct.

but `X` is not a free variable. the `X(r2)` claim is routed upstream via virtualization & eventually checked against model input. if the prover uses `X'(r2) != X(r2)`, the upstream chain rejects it. the prover can't change `X` w/o being caught.

the freedom you are looking for doesn't exist because the input `X` is pinned by upstream virtualization chain - not just the softmax protocol itself

AntoineF4C5:
@Forpee 

> So a malicious prover can't pick an X' thats inconsistent with the actual input - it would disagree with an evaluation somewhere along the chain.

In my view, the prover can in fact chose any input `Input'` for which he wants the verifier to believe `Model(Input') = Output`, and then the prover gets the trace, up to the softmax node's input `X'`.
At this point he also takes the real input `Input` such that `Model(Input) = Output`, and gets the trace **starting from** the softmax node (except the input `X`).

He can then use point 3 (and maybe point 1) to get the verifier to accept that on the softmax node, `X'` indeed leads to `Y`, and once we put everything together, prove that `Model(Input') = Output`.

Forpee:
@AntoineF4C5 - thx for the soundness review. Added a complementary-slackness sumcheck in Stage 2:

```
sum_{k,j} eq(r1, k,j) * sat_diff(k, j) * (z_bound - 1 - z_hi(k,j) * B - z_lo(k,j)) == 0
```

this forces `sat_diff = 0` when `z_c < z_bound − 1` (unsaturated), and only allows `sat_diff > 0` when `z_c = z_bound − 1` (saturated). So `z_c = min(z, z_bound − 1)` is uniquely determined

Bonus: the sumcheck naturally outputs `sat_diff(r_2)`, z_hi(r_2), `z_lo(r_2)` at the Stage 2 challenge point, so the manual `cache_z / cache_sat_diff` steps are gone

Note: after #61 is completed we can clamp z using that.

AntoineF4C5 (approved):
LGTM!
The inv_sum case would need more investigation to ensure it is a real vulnerability, and we plan to compute sum_exp_q instead of sending as auxiliary later, so it is better to not spend more time on that.

Forpee:
Thanks - agreed on both fronts. We'll revisit inv_sum when #218 lands (computing exp_sum_q inside the protocol removes the auxiliary entirely) .

Issue #208, opened 2026-03-27
Adapt Raf-check Identity Polynomial to match i32


AntoineF4C5:
This is an alternative solution to issue #119.

Raf-check asserts *more or less* `Input(r) = Sum( eq(r_cycle, j) * ra(j, k) * Int(k) )`

Currently, Input is defined as a tensor of `i32`, but `Int` casts bit-vector `k` into its unsigned integer representation, leading to a mismatch whenever Input has a negative value.

A solution idea is to adapt `Int` so that it casts k to its i32 signed representation.

PR #220, opened 2026-04-13
feat: abs lookup

Forpee:
Builds on Antoines' earlier work.

Required for #208 & #192.

See [here](https://github.com/ICME-Lab/jolt-atlas/blob/b9824b925e35c92724b87ef3717c530e3de38255/joltworks/src/lookup_tables/unsigned_abs.rs#L147-L158) for design choice around handling `int::MIN` edge case.

### DIFF
--- a/joltworks/src/lookup_tables/neg_relu.rs
+++ b/joltworks/src/lookup_tables/neg_relu.rs
@@ -62,6 +62,11 @@ impl<const X_LEN: usize> PrefixSuffixDecompositionTrait<X_LEN> for NegReluTable<
         suffixes: &[SuffixEval<F>],
     ) -> F {
         let [one, suffix_neg_relu] = suffixes.try_into().unwrap();
+        // Safe to multiply prefix[MSB] * prefix[NotLowerWord]:
+        //   - prefix[NotLowerWord] is degree-0 in the MSB variable
+        //   - prefix[MSB] is degree-1 in the MSB variable and degree-0 in the lower bits
+        // so their product remains degree ≤ 1 in every variable,
+        // preserving the multilinear property.
         prefixes[Prefixes::Msb] * prefixes[Prefixes::NotLowerWord] * one
             + prefixes[Prefixes::Msb] * suffix_neg_relu
     }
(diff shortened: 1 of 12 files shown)


joltworks/src/lookup_tables/prefixes/nlw.rs
```diff
@@ -0,0 +1,87 @@
...
+            }
+            (Some(r_x), _) => {
+                let x_shift = XLEN - jj;
+                let y_shift = x_shift - 1;
+                let r_x = if j == (XLEN + 1) {
+                    F::zero()
+                } else {
+                    F::one() - r_x.into()
+                };
+                word += F::from_u64(1 << x_shift) * r_x;
```
AntoineF4C5: I believe it is better to chose an other name than r_x for this variable (maybe "not_rx"), to avoid confusion.

```suggestion
                let not_rx = if j == (XLEN + 1) {
                    F::zero()
                } else {
                    F::one() - r_x.into()
                };
                word += F::from_u64(1 << x_shift) * not_rx;
```

joltworks/src/lookup_tables/neg_relu.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    #[cfg(test)]
+    fn combine_test<F: JoltField>(
+        &self,
+        prefixes: &[PrefixEval<F>],
+        suffixes: &[SuffixEval<F>],
+    ) -> F {
+        let [one, suffix_neg_relu] = suffixes.try_into().unwrap();
+        prefixes[Prefixes::Msb] * prefixes[Prefixes::NotLowerWord] * one
+            + prefixes[Prefixes::Msb] * suffix_neg_relu
```
AntoineF4C5, line 71: Can you add a comment mentionning that we can compute the product of prefixes Msb and NotLowerWord because they use strictly different bits from the lookup indice (hence has degree 1 in each variable).
I'm afraid this could else be confusing to anyone trying to learn to create new tables from this file.

AntoineF4C5 (approved):
Great implementation, LGTM!

joltworks/src/lookup_tables/neg_relu.rs
Forpee: yeah I agree - good call

PR #223, opened 2026-04-17
Feat: optimize commit witness polynomials

ClankPan:
## Summary

This PR improves the performance of `commit_witness_polynomials`.

After profiling the commitment phase, I found that most of the CPU time was being spent in **kernel time**, while **user time accounted for only about 8%**. This suggested that the main bottleneck was not the computation itself, but excessive **thread context switching** caused by overly fragmented Rayon jobs.

To address this, I changed the parallelization strategy so that the two main components called inside `commit_witness_polynomials`:

- `polynomial_map`
- `commit_to_polynomials`

are parallelized **only at the polynomial / commitment level**, while disabling parallelism inside their internal subroutines.

As a result, the commitment phase became significantly faster.

## Performance

Overall, this change yields a multi-fold speedup in the commitment phase.

As one concrete example, when running Qwen, the commitment time improved from:

- **38.6s -> 8.7s** (**77.5% reduction**, or about **4.4x faster**)

## Motivation

The previous implementation used `par_iter` in multiple layers within the commitment phase. In practice, this caused Rayon jobs to become too fine-grained, which led to substantial scheduling overhead and frequent thread context switching.

Profiling showed that kernel time dominated the execution, which strongly suggests that the system was spending much more time managing threads than performing useful user-space computation.

## Implementation

The implementation introduces a mechanism to suppress `par_iter`-based parallelism in the code paths related to `commit_witness_polynomials`.

In other words:

- `polynomial_map` itself is still parallelized at the polynomial level
- `commit_to_polynomials` itself is still parallelized at the commitment level
- lower-level internal routines called inside these phases no longer introduce additional nested parallelism

~~This is admittedly a somewhat ad hoc solution, but at the moment I was not able to find a better alternative that works within the current structure. If there is a cleaner or more principled solution, I would be very happy to adopt it.~~

The implementation now takes a different form.

Instead of relying on `maybe-rayon`, this PR introduces a new crate, `perhaps-rayon`, and switches the relevant code in:

- `jolt-atlas-core`
- `atlas-onnx-tracer`
- `joltworks`

to use it while keeping the diff against the original implementations as small as possible.

The main idea is to preserve a Rayon-compatible API while allowing the caller to choose, at runtime, whether a given operation should execute in parallel or sequentially.

In particular, `perhaps-rayon` provides APIs such as:

- `par_iter(bool)`
- `par_iter_mut(bool)`
- `into_par_iter(bool)`

where the boolean determines whether the operation uses parallel execution or falls back to sequential execution.

This differs from `maybe-rayon`, where that choice is effectively made only at compile time.

Using this runtime-selectable facade makes it possible to suppress nested parallelism in the hot paths related to `commit_witness_polynomials` without requiring larger structural changes across the workspace.

## Documentation

To help review the relevant execution structure, I added:

- `docs/commit_witness_polynomials_call_graph.md`

Please refer to that document as well when reviewing this PR.

I also created a new `docs/` directory because there was previously no obvious place to put this kind of internal implementation/performance note. If there is a better location for this type of document, I would appreciate suggestions.

## Notes

This PR focuses specifically on improving the commitment phase by reducing nested and fragmented parallelism. It does not attempt to provide a general-purpose scheduling or parallelism framework; it is a targeted fix for the current bottleneck.

### DIFF
--- a/atlas-onnx-tracer/src/ops/einsum.rs
+++ b/atlas-onnx-tracer/src/ops/einsum.rs
@@ -2,6 +2,7 @@ use crate::{
     ops::{Einsum, Op},
     tensor::{Tensor, TensorError},
 };
+use common::parallel::par_enabled;
 use rayon::prelude::*;
 use std::collections::{HashMap, HashSet};
 use tract_onnx::prelude::tract_itertools::Itertools;
@@ -169,6 +170,7 @@ pub fn einsum_i32_with_i64_rebase(
 
     let output: Vec<i32> = cartesian_coord
         .par_iter()
+        .with_min_len(par_enabled())
         .map(|out_coord| {
             let out_partials: Vec<usize> = (0..inputs.len())
                 .map(|inp_idx| {
--- a/atlas-onnx-tracer/src/ops/mul.rs
+++ b/atlas-onnx-tracer/src/ops/mul.rs
@@ -1,3 +1,4 @@
+use common::parallel::par_enabled;
 use rayon::prelude::*;
 
 use crate::{
@@ -44,6 +45,7 @@ pub fn mult_i32_with_i64_rebase(
         output
             .par_iter_mut()
             .zip(rhs_expanded.data().par_iter())
+            .with_min_len(par_enabled())
             .for_each(|(o, r)| {
                 let prod: i64 = (*o as i64) * (*r as i64);
 
--- a/atlas-onnx-tracer/src/tensor/mod.rs
+++ b/atlas-onnx-tracer/src/tensor/mod.rs
@@ -5,10 +5,11 @@ use crate::utils::parallel_utils::IndexedParallelIterator;
 use crate::utils::{
     self,
     dims::copy_strided,
-    parallel_utils::{IntoParallelRefIterator, IntoParallelRefMutIterator, ParallelSliceMut},
+    parallel_utils::{
+        IntoParallelRefIterator, IntoParallelRefMutIterator, ParallelIterator, ParallelSliceMut,
+    },
     quantize,
 };
-use maybe_rayon::iter::ParallelIterator;
 use rand::{Rng, RngCore, distributions::Uniform, rngs::StdRng};
 use serde::{Deserialize, Serialize};
 use std::{
--- a/atlas-onnx-tracer/src/tensor/ops.rs
+++ b/atlas-onnx-tracer/src/tensor/ops.rs
@@ -5,6 +5,7 @@ use crate::{
     tensor::{Tensor, TensorType},
     utils::parallel_utils::{IntoParallelRefIterator, IntoParallelRefMutIterator},
 };
+use common::parallel::par_enabled;
 use maybe_rayon::iter::ParallelIterator;
 use std::collections::{HashMap, HashSet};
 #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
@@ -53,6 +54,7 @@ pub fn iff<
     // assert is boolean
     if !mask
         .par_iter()
+        .with_min_len(par_enabled())
         .all(|x| *x == T::one().unwrap() || *x == T::zero().unwrap())
     {
         return Err(TensorError::WrongMethod);
@@ -134,6 +136,7 @@ pub fn or<
 ) -> Result<Tensor<T>, TensorError> {
     if !b
         .par_iter()
+        .with_min_len(par_enabled())
         .all(|x| *x == T::one().unwrap() || *x == T::zero().unwrap())
     {
         return Err(TensorError::WrongMethod);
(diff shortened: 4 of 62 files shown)


wyattbenno777:
update the main readme.md with better times when they are achieved.

Forpee:
Hey @ClankPan - I haven't done a full review yet but wanted to flag a few high-level concerns early so we can align before this grows any further.

1. PR scope - please split this up
The original goal was to fix the commit witness parallelization strategy (and the numbers look great 38s->8s). But this PR has expanded well beyond that into a full codebase-wide replacement of rayon with a new `perhaps-rayon` library. That's a fundamentally different change from the targeted commit-witness optimization.

Can we split this into at least two PRs?

PR A: The targeted `commit_witness_polynomials` parallelization fix (the original intent -  `ParallelFlagGuard`, granularity tuning etc,...
PR B (separate, if we decide to pursue it): The `perhaps-rayon` abstraction and global rayon replacement

The commit-witness speedup is valuable and reviewable on its own. Bundling it with a 7,000-line new library and a global refactor makes it much harder to review, bisect, and reason about regressions

2. `perhaps-rayon` - maintenance burden & scope concerns
`perhaps-rayon` is a ~7,000-line rayon facade. That's a non-trivial library to own and maintain inside a proving system repo. Some concerns:

* Maintenance surface: We're now on the hook for keeping parity with rayon's iterator API (the README already has a long checklist of covered vs. uncovered surface).
* Scope: This is a general-purpose parallelism abstraction - it doesn't belong in a zk proving codebase. If we genuinely need this (and I'm not yet convinced we do), it should live in its own repo with its own CI and versioning.
* Alternatives: This PR already introduces ParallelFlagGuard in common::parallel, which provides exactly the scoped sequential/parallel toggling needed for the commit-witness fix. That's zero new dependencies, and it solves the granularity problem cleanly. It's unclear what `perhaps-rayon` (~7,000 lines, a full rayon iterator facade) adds on top of that. Other proving systems facing the same rayon granularity challenges (e.g., a16z/jolt, our upstream reference) use similarly lightweight scoped controls without wrapping rayon's entire API. 

3. Missing e2e benchmarks
The PR description cites the commit_witness_polynomials speedup ("38.6s -> 8.7s"), which is solid. But the global `perhaps-rayon` migration touches hot paths across the entire prover — `joltworks` sumcheck, polynomial operations, MSM, lookup tables, etc. What's the e2e proving time before and after this PR? Without that number we can't tell if the rayon replacement is neutral, beneficial, or a regression outside of the commit-witness path

tl;dr: The commit-witness optimization is great - let's land that on its own. The `perhaps-rayon` migration is a much bigger conversation that deserves its own PR, benchmarks, and design discussion

ClankPan:
@Forpee 
Thank you for the feedback.

I agree that `perhaps-rayon` is a fairly large change. Initially, for all `par_iter` usages related to `commit_witness_polynomial`, I had added branching like this:

```rust
if par_enabled() {
  vec.par_iter().map(|item| ...).collect()
} else {
  vec.iter().map(|item| ...).collect()
}
```

However, the relevant `par_iter` calls are scattered across `joltworks`, `onnx-tracer`, and `jolt-atlas-core`, and adding these `if-else` branches everywhere would significantly hurt readability and make the diff much noisier.

That is why I took this approach instead, in order to keep the code changes as small as possible. The implementation is [here](https://github.com/ICME-Lab/jolt-atlas/compare/main...wip/optimize-commit).

If you think the alternative approach would be more maintainable, I’m happy to revert to that instead.

ClankPan:
For now I’ll focus on extracting the original optimization into a minimal PR first and I’ll split this up. 

Plan:
- PR A: the targeted `commit_witness_polynomials` optimization only
  - scoped parallel suppression via `ParallelFlagGuard`
  - limiting parallelism granularity in `polynomial_map` / `commit_to_polynomials`
  - the call graph doc and the localized benchmark numbers
- PR B: the `perhaps-rayon` abstraction as a separate discussion, if we still think it is worth pursuing

I also agree that for PR B I need to justify the maintenance cost much more clearly and provide end-to-end proving benchmarks, not just the localized commitment-phase speedup.

ClankPan:
@Forpee 
I found a way to suppress Rayon parallelism at runtime using a `with_min_len`-based hack. If `usize::MAX` is passed, the iterator no longer splits its work, so it is effectively processed sequentially. If `1` is passed, the minimum split size is 1, which corresponds to the normal behavior.

This lets us keep the diff minimal, and I have not observed any performance regression from this change.

I would appreciate further review.

Forpee (approved):
Excellent work on identifying the parallelization bottleneck. Let's address the ai-comments (duplicate `with_min_len` calls, `zip + with_min_len`, etc..) and update the readme with the benchmark results.

Performance improvement confirmed: gpt2 commitment time reduced from ~4s to ~760ms with no observed regressions elsewhere.

The root cause analysis is spot-on - the nested par_iter() calls in the commitment path were creating excessive scheduling overhead. Coarsening the granularity to the polynomial/commitment level was exactly the right fix.

ClankPan:
@Forpee 
Thank you for the review. I fixed them by adding `fn par_enabled_with(len: usize)`.
About the benchmark documents, let me make the benchmark changes in a separate PR. There are still a few outstanding issues, such as the validation items and running Qwen. Including those would go beyond the scope of this PR.

PR #226, opened 2026-04-20
feat: signed-raf

### LINKED ISSUE #208: Adapt Raf-check Identity Polynomial to match i32

Forpee:
closes #208

- [x] signed-identity-poly
- [ ] signed operand poly
- [ ] wire up to jolt-atlas-core

### DIFF
--- a/joltworks/src/field/ark.rs
+++ b/joltworks/src/field/ark.rs
@@ -97,6 +97,16 @@ impl JoltField for ark_bn254::Fr {
         }
     }
 
+    #[cfg(test)]
+    fn from_i8(val: i8) -> Self {
+        if val.is_negative() {
+            let val = val.unsigned_abs();
+            -<Self as JoltField>::from_u8(val)
+        } else {
+            <Self as JoltField>::from_u32(val as u32)
+        }
+    }
+
     #[inline]
     fn from_i32(val: i32) -> Self {
         if val.is_negative() {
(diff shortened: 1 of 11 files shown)


AntoineF4C5:
I believe there is a significantly easier way to implement this.
The current way to prove `Raf` is to multiply the on-hot encoded tensor to an identity tensor, and assert the output equals the `Raf` value.
 
```example
x=5, n= 3, N = 2^n = 8
x_oh = [0, 0, 0, 0, 0, 1, 0, 0]
Int =  [0, 1, 2, 3, 4, 5, 6, 7]
x_oh · Int = 1 × 5 = x
```
This is encoded with a prefix-suffix decomposition essentially representing (with `ẋ` the bit of bitwise decomposition of x):
```example
Int(ẋ) = ∑_{0≤i<n} x^(n-i-1) * ẋ_i
```
This considers an Unsigned encoding of numbers.
Here is the prefix part represented here: https://github.com/ICME-Lab/jolt-atlas/blob/bee497838fa8c0ec844fa7867e66708d4adf69b5/joltworks/src/poly/identity_poly.rs#L145-L147

To have our identity follow the 2's complement writing, we just need to slightly adapt the writing:
```example
Int(ẋ) = -2^n × ẋ_0 + ∑_{1≤i<n} x^(n-i-1) * ẋ_i
```

To implement Prefix-Suffix decomposition on that, we can see we do not need to modify the suffix part, since the lower bits are handled the same way.
The only thing that needs to differ is the "coefficient" of the MSB/sign-bit, which needs to change from `2^(n-1)` to `-2^n`.
I hope there is no caveat I missed.

PR #229, opened 2026-04-23
Add verifier checks

Acentelles:
This PR checks the output claim is bound to the computation and that Fiat-Shamir challenges were correctly derived.

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/zk.rs
+++ b/jolt-atlas-core/src/onnx_proof/zk.rs
@@ -517,7 +517,8 @@ pub fn verify_zk(
     let output_node = &model[output_index];
     let r_node_output = transcript
         .challenge_vector_optimized::<F>(output_node.pow2_padded_num_output_elements().log_2());
-    let r_node_output_field: Vec<F> = r_node_output.iter().map(|c| (*c).into()).collect();
+    use joltworks::field::IntoOpening;
+    let r_node_output_field: Vec<F> = r_node_output.into_opening();
     let expected_output_claim =
         MultilinearPolynomial::from(io.outputs[0].padded_next_power_of_two())
             .evaluate(&r_node_output_field);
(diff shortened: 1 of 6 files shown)


jolt-atlas-core/src/onnx_proof/zk.rs
```diff
@@ -397,23 +476,133 @@ pub fn prove_zk(
...
+    for commitment in &bundle.commitments {
+        transcript.append_serializable(commitment);
+    }
+
+    // 2. Verify output claim: derive challenge, evaluate output MLE, check match
+    let output_index = model.outputs()[0];
+    let output_node = &model[output_index];
+    let r_node_output = transcript
+        .challenge_vector_optimized::<F>(output_node.pow2_padded_num_output_elements().log_2());
+    let r_node_output_field: Vec<F> = r_node_output.iter().map(|c| (*c).into()).collect();
```
AntoineF4C5: This may not be so helpful, but I added a helper to get the Field Vector from Challenge vector.

```suggestion
    use joltworks::field::IntoOpening;
    let r_node_output_field: Vec<F> = r_node_output.into_opening();
```

jolt-atlas-core/src/onnx_proof/zk.rs
```diff
@@ -129,42 +134,85 @@ pub fn prove_zk(
...
+                .insert(node_idx, reduced);
+            eval_reduction_proofs.insert(node_idx, proof);
+            if let Some(com) = h_com {
+                eval_reduction_h_commitments.insert(node_idx, com);
+            }
+        }
 
-        // ZK sumcheck based on operator type
-        match &node.operator {
+        let zk_proof = match &node.operator {
```
AntoineF4C5: This "match" case executes `run_zk_sumcheck` for all sumcheck operators.
Can we use a method analogous to "create_verifier_instances" that outputs a SumcheckProver to deduplicate?

AntoineF4C5 (approved):
There's a few compiler warnings with the `zk` feature.
I also made a few comments.

Other than that, LGTM!

PR #244, opened 2026-05-26
fix(zk): bind public-node Constant/Input openings via BlindFold R1CS

Acentelles:
The cleartext `public_node_reduced_claims` check in `verify_zk` was decoupled from the BlindFold witness, leaving the Constant/Input opening values in the witness unbound to the public tensors of the verifier's model. A malicious prover could:

  1. Run the honest ZK prover on a different model M' (e.g. with a different Constant tensor than the verifier's M).
  2. Patch `bundle.public_node_reduced_claims` to the values M's verifier expects (a deterministic function of the public M tensor, so the attacker can compute them).
  3. Submit. The cleartext check matches the patched claims, the BlindFold witness's Constant/Input openings are unbound, and the verifier accepts a proof of `y = input + M'.constant` against M.

Fix: for every `NodeExecution` opening on a `Constant` or `Input` node, add an R1CS extra-constraint to the BlindFold witness binding the opening value to the public-MLE evaluation at the opening point.

  Constraint:        output_var = Opening(id) - Challenge(0)
  Baked challenge:   public_mle(opening_point)  (deterministic, public)
  Witness:           output_value = 0, blinding = 0
  y_com:             identity G1

Verifier side:
  1. Independently compute `expected = public_mle_M(opening_point)`.
  2. Check `baked.extra_constraint_challenges[i] == expected`.
  3. Check `eval_commitments[i]` is identity G1.
  4. BlindFold's `eval_commitment_gens` check enforces `y_com == g*output_var + h*blinding`; identity ⇒ output_var = 0 AND blinding = 0.
  5. The R1CS row then forces `Opening = Challenge = expected`.

Filter: only `SumcheckId::NodeExecution` openings (consumer-side raw NodeOutput evaluations) are bound. Internal protocols (`Raf`, `RaVirtualization`, `RamHammingBooleanity`, ...) open transformed representations under the same polynomial-id; their witness values aren't raw MLE evals and a binding to `public_mle(opening_point)` would (correctly) reject. Those internal openings remain bound by their own protocol-specific algebraic constraints.

Regression test `test_zk_constant_binding_rejects_cross_model_attack` exercises both layers: (a) unpatched malicious bundle rejected by the cleartext check; (b) bundle patched with verifier-expected `public_node_reduced_claims` rejected by the new R1CS binding with `Public-node binding mismatch`. A test-only helper `extract_verifier_expected_public_claims` runs the verifier flow once to recover the expected values so the test can construct the post-cleartext-patch attack.

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/e2e_tests.rs
+++ b/jolt-atlas-core/src/onnx_proof/e2e_tests.rs
@@ -843,6 +843,139 @@ fn test_add_zk() {
         .expect("ZK verification should succeed");
 }
 
+/// Regression test for the `Constant`/`Input` opening-binding fix.
+///
+/// Setup: build two models M and M' with identical topology
+/// (`y = input + constant`) but different `Constant` tensors. A malicious
+/// prover runs the honest ZK prover on M' (so its BlindFold witness
+/// encodes M''s constant value) and submits the resulting bundle to a
+/// verifier configured for M. Two layers of defense are exercised:
+///
+/// 1. The cleartext `public_node_reduced_claims` check rejects the bundle
+///    as-is, because the bundle's claims are computed against M' but the
+///    verifier expects M's.
+/// 2. **The interesting test**: even if the attacker patches
+///    `public_node_reduced_claims` to match the verifier's expected
+///    values (which an attacker can do — it's a deterministic function
+///    of the bundle and the public M tensor), the R1CS binding
+///    constraints added in this commit reject the attempt. The binding
+///    constraints tie each `Opening(NodeOutput(constant), NodeExecution(consumer))`
+///    in the BlindFold witness to `public_mle(opening_point)` via a
+///    public Challenge slot and an identity-G1 y_com. The verifier
+///    independently computes the expected challenge values and compares
+///    them against `bundle.baked.extra_constraint_challenges`; on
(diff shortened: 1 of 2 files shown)


omarespejel:
Hi @Acentelles @wyattbenno777 @AntoineF4C5!

I re-tested the original `verify_zk` input-rebinding repro from the advisory i opened and this PR’s merge commit `684bc745a7b171ff1291d422d106bf384fce3aeb` now rejects the same `square(input)` `x -> -x` case that previously passed in the optional ZK verifier

From my side, PR #244 looks like the fix for the advisory

Could we link this PR from the advisory and publish or resolve the advisory while keeping the reporter credit visible (would love also if we can do an X post 🙏)? Its helpful for my verifiable zk work and open source contributions. Happy to help check the final wording!

Thanks again

Forpee:
Hi @omarespejel 
Can u send a link for the advisory. I can't seem to find it

wyattbenno777:
Hey @Forpee ,

Its hidden by default for security reasons. 

Here is the published version:
https://github.com/ICME-Lab/jolt-atlas/security/advisories/GHSA-333f-w2cf-wch9

@omarespejel sounds good! Let us know about a tweet or blog we can help boost it.

PR #245, opened 2026-05-29
signed-raf

### LINKED ISSUE #195: Soundness: teleport range checks do not bind tau-derived bounds to the verifier
## Summary

  Tanh/Erf neural-teleport range checks appear to accept a prover-chosen tau-derived bound, instead of forcing the verifier to bind that
  bound to the operator's actual tau.

  This means the prover may be able to prove a different statement from the intended one:

  - intended: remainder < tau
  - actually accepted: remainder < tau'

  ## Why this matters

  In the teleport flow, tau is used for:

  - the division relation
  - the lookup flow
  - the remainder range check

  The range-check side should be tied to the same tau as the operator semantics.

  Right now, there seems to be a gap: the range-check bound tensor can be changed by the prover without the verifier independently
  rejecting it.

  ## PoC

  I added a soundness regression test for Tanh to [this branch](https://github.com/ICME-Lab/jolt-atlas/tree/exp/tau-rangecheck-bypass-poc):

  - soundness_tanh_tau_rangecheck_bypass_is_rejected

  The malicious prover:

  - keeps Stage 1a division honest
  - keeps Stage 1b lookup honest
  - changes only the Stage 2/3 range-check path to use tau + 1
  - also changes the TeleportRangeCheckRaD witness to stay consistent with that forged bound

  The honest verifier should reject this proof.

  At the moment, the proof still verifies, so the regression test is marked #[ignore] and fails when run.

  ## Expected behavior

  The verifier should reject any proof where the range-check bound is not the one implied by the operator's real tau.

  ## Suggested fix direction

  The RAF right-operand claim used by teleport range checking should be explicitly bound to the verifier-derived tau tensor, including
  padded indices, rather than accepted only through prover-supplied virtual claims.

  ## Reproduction

  Run:

  `cargo test -p jolt-atlas-core soundness_tanh_tau_rangecheck_bypass_is_rejected -- --ignored`

  If the issue is present, the test fails because the malicious proof is accepted.

### LINKED ISSUE #208: Adapt Raf-check Identity Polynomial to match i32

Forpee:
Implements prefix-suffix decomposition for the signed-identity MLE:

`IntSigned(r) = sum_{i=0}^{n-1} r[n-1-i] * 2^i - r[0] * 2^n`

This completes the read-raf protocol for our signed use cases. The RAF protocol now uses the signed identity instead of the unsigned identity, . enabling removal of the dummy RAF claims. which was previously causing soundness bugs

Future work: implement a unary-specific prefix-suffix-shout that sums over 32 variables instead of the 64-variable scheme used for binary ops. This would remove the need for `BinarySignedIdentityPoly`, which treats the first `X_LEN` variables as `0`, and would simplify single-operand operators like relu while reducing the sum-check rounds by 32.


jolt-atlas-core/src/onnx_proof/neural_teleport/reduction.rs
```diff
@@ -0,0 +1,244 @@
...
+    subprotocols::{
+        sumcheck_prover::SumcheckInstanceProver,
+        sumcheck_verifier::{SumcheckInstanceParams, SumcheckInstanceVerifier},
+    },
+    transcripts::Transcript,
+};
+
+/// Shared parameter block for the reduction sumcheck proof.
+#[derive(Clone)]
+pub struct ReductionParams<F: JoltField> {
```
AntoineF4C5: I'm thinking of finding an other name for this protocol, as we are shifting the evaluation point of the claim from an existing `r` to a new point `r'` recovered during the sumcheck protocol, rather than "reducing" several claims into one.
Using an other name could avoid confusion with the claim reduction protocol.

Proposed name: `EvalShifting`

jolt-atlas-core/src/onnx_proof/ops/erf.rs
```diff
@@ -213,11 +220,15 @@ impl<F: JoltField> SumcheckInstanceParams<F> for ErfParams<F> {
...
+            SumcheckId::NTReduction,
+        );
+        let rv_claim = accessor.get_custom(reduced_output_id).1;
 
         let quotient_id = OpeningId::new(
             VirtualPoly::TeleportQuotient(self.computation_node.idx),
-            SumcheckId::Raf,
+            SumcheckId::NodeExecution(self.computation_node.idx),
         );
         let quotient_claim = accessor.get_custom(quotient_id).1;
```
AntoineF4C5: Since the SumcheckId is now `NodeExecution(node_idx)` variant, we can use the `get_advice` accessor method to recover the quotient claim.

```suggestion
        let quotient_claim = accessor.get_advice(VirtualPoly::TeleportQuotient).1;
```

jolt-atlas-core/src/onnx_proof/ops/sigmoid.rs
```diff
@@ -213,11 +220,15 @@ impl<F: JoltField> SumcheckInstanceParams<F> for SigmoidParams<F> {
...
+            SumcheckId::NTReduction,
+        );
+        let rv_claim = accessor.get_custom(reduced_output_id).1;
 
         let quotient_id = OpeningId::new(
             VirtualPoly::TeleportQuotient(self.computation_node.idx),
-            SumcheckId::Raf,
+            SumcheckId::NodeExecution(self.computation_node.idx),
         );
         let quotient_claim = accessor.get_custom(quotient_id).1;
```
AntoineF4C5: Same as Erf.

```suggestion
let quotient_claim = accessor.get_advice(VirtualPoly::TeleportQuotient).1;
```

jolt-atlas-core/src/onnx_proof/ops/tanh.rs
```diff
@@ -214,12 +217,16 @@ impl<F: JoltField> SumcheckInstanceParams<F> for TanhParams<F> {
...
+        );
+        let rv_claim = accessor.get_custom(reduced_output_id).1;
 
         // Use quotient claim instead of input claim (neural teleportation)
         let quotient_id = OpeningId::new(
             VirtualPoly::TeleportQuotient(self.computation_node.idx),
-            SumcheckId::Raf,
+            SumcheckId::NodeExecution(self.computation_node.idx),
         );
         let quotient_claim = accessor.get_custom(quotient_id).1;
```
AntoineF4C5: Same as Erf.

```suggestion
let quotient_claim = accessor.get_advice(VirtualPoly::TeleportQuotient).1;
```

joltworks/src/field/ark.rs
```diff
@@ -99,6 +99,16 @@ impl JoltField for ark_bn254::Fr {
         }
     }
 
+    #[cfg(test)]
+    fn from_i8(val: i8) -> Self {
+        if val.is_negative() {
+            let val = val.unsigned_abs();
+            -<Self as JoltField>::from_u8(val)
+        } else {
+            <Self as JoltField>::from_u32(val as u32)
```
AntoineF4C5: Just of my curiosity: Why not using from_u8 as for the positive case ?

joltworks/src/poly/multilinear_polynomial.rs
```diff
@@ -683,10 +683,18 @@ pub trait PolynomialEvaluation<F: JoltField> {
     where
         Self: Sized,
         C: Copy + Send + Sync + Into<F> + ChallengeFieldOps<F>,
-        F: FieldChallengeOps<C>;
+        F: FieldChallengeOps<C>,
+    {
+        let _ = (polys, r); // to silence unused variable warnings
+        unimplemented!("Implemented batch evaluation for MultilinearPolynomial");
```
AntoineF4C5: I do not get the sense of this.

joltworks/src/subprotocols/ps_shout.rs
```diff
@@ -233,33 +269,30 @@ where
...
-        let T = 1 << params.log_T;
-        let phases = 8;
+    fn new_inner(
+        params: ReadRafSumcheckParams<F, LUT, SIGNED>,
+        lookup_indices: Vec<LookupBits>,
+        identity_ps: PrefixSuffixDecomposition<F, 2, SIGNED>,
+        left_operand_ps: PrefixSuffixDecomposition<F, ORDER_OP, false>,
+        right_operand_ps: PrefixSuffixDecomposition<F, ORDER_OP, false>,
+    ) -> Self {
+        let phases = NUM_PHASES;
```
AntoineF4C5, line 277: Here the `identity_ps` reuses the "SIGNED" const generics, but left/right_operand_ps uses hardcoded "false".

AntoineF4C5 (commented):
Nice work ! Cool to see everything starting to get linked together.

I left some comments on questions I had, things I noticed.

Can you sumarize where are we still using unsigned identity for raf-checking (or any other use)? I'm wondering if we could not simply get rid of the unsigned identity, and remove the complexity of the `SIGNED` const generics for prefix-suffix decomposition.

joltworks/src/poly/multilinear_polynomial.rs
Forpee: the goal was to reduce boilerplate for polys that need the `evaluation` method from `PolynomialEvaluation` but don't need to implement `batch_evaluate` or `sumcheck_evals`.

Can u clarify what u don't understand? In any case if its what i described, i can remove it and move the boilerplate to the implementation sites

jolt-atlas-core/src/onnx_proof/neural_teleport/reduction.rs
Forpee: The naming stems from here: https://github.com/ICME-Lab/jolt-atlas/blob/684bc745a7b171ff1291d422d106bf384fce3aeb/joltworks/src/subprotocols/opening_reduction.rs#L50

Where the sum-check reduces the claim `Y(r)` to a new claim `Y(r')`.
That said I agree theres potential for confusion with the other eval-reduction-protocol, so im happy to rename it

Forpee:
> Nice work ! Cool to see everything starting to get linked together.
> 
> I left some comments on questions I had, things I noticed.
> 
> Can you sumarize where are we still using unsigned identity for raf-checking (or any other use)? I'm wondering if we could not simply get rid of the unsigned identity, and remove the complexity of the `SIGNED` const generics for prefix-suffix decomposition.

@AntoineF4C5 I added `SIGNED` originally to preserve backwards compatibility, but you're right we don't use unsigned-raf for any of our prefix-suffix-shout instances.

The only cases where we use unsigned-raf are when indices are positive, like gather or softmax-exponentiation. But those are not prefix-suffix-shout instances just regular ones.

So Im okay with simplifying the protocol, rm `SIGNED` and hardcode it to use signed-raf

jolt-atlas-core/src/onnx_proof/ops/erf.rs
Forpee: clean!

joltworks/src/field/ark.rs
Forpee: We can use `from_u8` - thanks for spotting it.

joltworks/src/subprotocols/ps_shout.rs
Forpee: good catch!

joltworks/src/poly/multilinear_polynomial.rs
AntoineF4C5: Maybe it's just the sentence used in the "unimplemented!" macro that I did not understand then.
Are there a lot of such polynomials ? The potential issue with default implementation is that implementors could forget to reimplement them in some cases, so if this is only 1 or 2 polynomials that do not require batch_evaluate and sumcheck_evals, I'm happy to move the boilerplate to implementation!

jolt-atlas-core/src/onnx_proof/neural_teleport/reduction.rs
AntoineF4C5: Got it.
If my understanding is correct, in opening_reduction.rs there are a still several opening points, and the protocol reduces them to 1, or am I wrong ? (Trying to make sense of using "reduction" for that other protocol)

AntoineF4C5:
> So Im okay with simplifying the protocol, rm SIGNED and hardcode it to use signed-raf

@Forpee Nice!
I believe simplifying can only benefit us in the long run.
Also, I forgot to mention it but since we are now using the SignedIdentityPoly for teleported operators, do we still use "TeleportIdPolynomial" ? Maybe we could also remove it as it is kind of a less-optimized duplicate of the SignedIdentityPoly

joltworks/src/poly/multilinear_polynomial.rs
Forpee: That sentence is just the panic message that gets displayed in terminal if the default impl is called.

The idea was that if someone forgets to implement it but ends up needing it, the prover would panic & make the missing impl obvious.

That said your concern makes sense. Only a small number of polys require this. I'll remove the default impl and move the boilerplate to those impls instead

jolt-atlas-core/src/onnx_proof/neural_teleport/reduction.rs
Forpee: not exactly. It reduces multiple openings at different points to multiple openings at the same point.

But agreed the naming can be confusing here. To avoid bikeshedding I'll change it

Issue #230, opened 2026-04-28
Soundness bug of Input evaluation


AntoineF4C5:
In our current implementation, we do not commit to the `Input` tensors of the model.
As I was fastly reviewing, I realized we are also never binding the transcript to the `Input`.
Which means a malicious prover can decide of an `Output` he wants to claim is recovered from forged `Input'` the following way:
- First identify `Input` such that `Model(Input) = Output`.
- Trace model with `Input` inputs.
- Prove the model with `Input`.
  We are particularly interested in the proving of `Input` nodes, where the verifier just recovers the challenge evaluation point for this node output (Which is a model `Input`), and evaluates the tensor at given point.
- The prover recovers all (evaluation claim, evaluation point) tuples for `Input` of the model, and can now modify its inputs to any tensors of his liking, as long as their evaluation at `evaluation point` equals `evaluation claim`.
- The prover then sends the proof obtained in step 3, and just changes the `Input` part of the proof to `Input'`.

This works because we are not updating the transcript with the input values, so modifying the model's inputs doesn't change the challenges derived (as long as everything else in the Trace stays the same).

This can be fixed by having the protocol append `Input` or a hash of it to the transcript at the beginning of proving.
This issue would also be fixed by #201

PR #247, opened 2026-06-04
fix(soundness): bind public inputs into the transcript (#230)

Acentelles:
Fix: absorb the public input tensors (node index, shape, values) into the transcript before any challenge is derived, symmetrically on prover and verifier, in both the standard and zk paths. Any change to `io.inputs` now diverges the verifier's transcript from the proof and the IOP rejects before the evaluation check is reached, so the opening points can no longer be predicted and there is no fixed target to craft a colliding `Input'` against.

- mod.rs: `append_inputs_to_transcript`; called in `ONNXProof::prove`/`verify`.
- zk.rs: called in `prove_zk` and `verify_zk_with_pcs_capture_and_extract`.
- malicious_prover.rs: both variants mirror the absorption (a faithful attacker follows the protocol; only the forgery deviates) so the soundness tests still exercise their intended defense rather than transcript divergence.
- e2e_tests.rs: rebound-input regression now caught by transcript binding; new `test_input_binding_changes_transcript` unit test.

Outputs are already bound (output-claim check + IOP), so only inputs need binding here.

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/e2e_tests.rs
+++ b/jolt-atlas-core/src/onnx_proof/e2e_tests.rs
@@ -976,6 +976,100 @@ fn test_zk_constant_binding_rejects_cross_model_attack() {
     );
 }
 
+/// Regression for issue #230 (the input-binding soundness bug).
+/// `verify_zk` once accepted a proof for `output =
+/// square(input)` when `ModelExecutionIO.inputs[0]` was rebound from `x` to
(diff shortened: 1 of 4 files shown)


Acentelles:
Addresses: https://github.com/ICME-Lab/jolt-atlas/issues/230

antoine-af (approved):

AntoineF4C5 (approved):
LGTM!

PR #249, opened 2026-06-11
chore: refactor prefix checkpoints handling

AntoineF4C5:
Refactors the structure used to hold checkpoint values for each Prefix of the Prefixes enum, used to prove table lookups. 
Concretely, the new structure now allows to have a consistent behavior when indexing the checkpoint array by `usize` and `Prefixes`.

This refactor was conducted while implementing #250, to allow defining a generic Prefix, and have several implementations with different concrete behavior.

### DIFF
--- a/joltworks/src/lookup_tables/prefixes/mod.rs
+++ b/joltworks/src/lookup_tables/prefixes/mod.rs
@@ -180,13 +180,13 @@ impl<F: JoltField> Index<usize> for PrefixCheckpoints<F> {
     type Output = Option<F>;
 
     fn index(&self, index: usize) -> &Self::Output {
-        &self.0.get(index).unwrap().0
+        &self.0[index].0
     }
 }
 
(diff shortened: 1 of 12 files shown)


joltworks/src/lookup_tables/prefixes/mod.rs
```diff
@@ -223,8 +274,9 @@ impl Prefixes {
         F: JoltField + FieldChallengeOps<C>,
     {
         debug_assert_eq!(checkpoints.len(), Self::COUNT);
```
Forpee: this check is now redundant

Forpee (approved):
LGTM. This gives us a dedicated container for the full checkpoint set, indexable as `checkpoints[i]` or `checkpoints[prefix]` instead of arbitrary slice. Tradeoff is we lose slice methods like `.to_vec()`, `iter()`, `.iter_mut()`. unless we add them back. Also note `checkpoints[prefix]` returns `Option<F>` not `PrefixCheckpoint<F>`

PR #250, opened 2026-06-11
feat: Clamp generic lookup table and allow various implementations

### LINKED ISSUE #61: Implement clamping instruction

AntoineF4C5:
## Clamp lookup table

Implements the Clamp lookup table and its prefix-suffix decomposition.
The lookup table admits a generic const `BOUND` which represent the bit-index of high bound of the clamping to be performed.
Hence the input gets clamped in the range `[-2^BOUND; 2^BOUND)`.
It also admits a non-symmetric mode, where the input is clamped to [0, 2^BOUND).

This table is implemented for:
- Saturating some results to the i32 range, for operators whose output may exceed this range, such as Einsum.
- Reducing the domain of activation operators, for which the output is bounded by [-1; 1], and these bound values are fastly reached with inputs outside, for example, [-4; 4).
- Reducing the domain of exponentiation in Softmax, which we manipulate to always compute the exponent of negative values, with `e^(-x)` being close to zero as x grows.

#### Activation operators bench

| Variant | Prove (avg) | Verify (avg) |
|---|---|---|
| Baseline (neural_teleport, log_table=16) | 3.34s| 7.37ms |
| Clamped small table (log_table=12) | 3.33s | 6.56ms |

#### Softmax operator bench
| Variant                                       | Scale | Prove   | Verify |
| --------------------------------------------- | ----- | ------- | ------ |
| Baseline (`sat_diff`)                         | 8     | 63.31ms | 4.48ms |
| Clamped (`ClampBoundedTable`) | 8     | 71.87ms | 5.07ms |
| Baseline (`sat_diff`)                         | 12    | 69.41ms | 5.17ms |
| Clamped (`ClampBoundedTable`) | 12    | 78.43ms | 5.51ms |

We can reduce the proving time by lowering the table size, if we are guaranteed the inputs of the table already feat in any range lower than i32. (which was already assumed and optimized against in previous activation/softmax implementation.)

Should hopefully be easier to understand the PR by looking commit after commit.

### Additional changes
Added a new test method for lookup tables, to ensure the so-called multilinear extension defined in the implementation is indeed multilinear.

---

Closes #61


Forpee:
https://github.com/ICME-Lab/jolt-atlas/blob/6ef6addf8accf4a1a352c702487acbe1dc446107/joltworks/src/lookup_tables/test.rs#L54
lovely addition

Forpee:
I'd prefer tanh not to be so rigidly coupled to the default scale (as we need scale=12 for accurate gpt2).  Is the plan to address this in a follow-up? 

The same applies to cos and sin, which we also plan to support different scales?

AntoineF4C5:
@Forpee The different operators on which clamp is implemented are coupled to scale since the range at which we want to clamp depends on the scale. However the implementation is still compatible with any scale, the choice made here is to make the MODEL_SCALE a compile time constant, and consider we compile the VM with the desired scale set.

Let us assess if we want the VM to support all scales from a unique binary, and the costs of it.

jolt-atlas-core/src/onnx_proof/ops/activation_clamped/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+    fn r_cycle_source(node_idx: usize) -> OpeningId {
+        DefaultLookupOperands::r_cycle_source(node_idx)
+    }
+
+    fn witness(&self, node: &ComputationNode, trace: &Trace) -> Tensor<i64> {
+        DefaultLookupOperands.witness(node, trace)
+    }
+
+    fn lookup_bits(witness: &Tensor<i64>) -> Vec<LookupBits> {
+        let operand = witness.map(|v| v as i32 + OFFSET);
```
Forpee: `v as i32 + OFFSET` overflows (panic) for any input above `i32::MAX - 2^BOUND`. We need a 64-bit lookup index like `SaturationTable` uses or or a two-sided table that floors at `-2^B` directly with no offset - `SatClampTable` already does exactly that via `upper_eqo` alongside `upper_eqz`, just hardwired to a half-word split rather than an arbitrary bound

jolt-atlas-core/src/onnx_proof/witness.rs
```diff
@@ -309,6 +275,54 @@ impl<F: JoltField> WitnessGenerator<F> for CommittedPoly {
...
+                let computation_node = &model.graph.nodes[node_idx];
+                let layer_data = Trace::layer_data(trace, computation_node);
+                let padded_operands: Vec<_> = layer_data
+                    .operands
+                    .iter()
+                    .map(|tensor| {
+                        tensor
+                            .padded_next_power_of_two()
+                            .par_enum_map(|_, v| -> Result<i32, TensorError> {
+                                Ok(v + SymmetricClampOperands::OFFSET)
```
Forpee: ditto on the panic

jolt-atlas-core/src/onnx_proof/witness.rs
```diff
@@ -309,6 +275,54 @@ impl<F: JoltField> WitnessGenerator<F> for CommittedPoly {
...
+                let computation_node = &model.graph.nodes[node_idx];
+                let layer_data = Trace::layer_data(trace, computation_node);
+                let padded_operands: Vec<_> = layer_data
+                    .operands
+                    .iter()
+                    .map(|tensor| {
+                        tensor
+                            .padded_next_power_of_two()
+                            .par_enum_map(|_, v| -> Result<i32, TensorError> {
+                                Ok(v + (1 << ACTIVATION_BOUND))
```
Forpee: ditto

Forpee (approved):
* Core impl is correct.
* Flexible and generic Clamping table is a good abstraction.
* Activation_clamped pipeline is clean.
* Softmax nicely simplified.

Just left a comment regarding overflows in the witness compution

Separately flagging: gpt2 CI is red. How would you like to proceed?

jolt-atlas-core/src/onnx_proof/ops/activation_clamped/mod.rs
AntoineF4C5: Indeed, I had not thought of this caveat. The SatClamp table is probably best to use for that reason.
I will make it generic over the bound choice, to use it for activation too.

AntoineF4C5:
@Forpee I fixed the issues you numbered and got the gpt2 test passing when I fix the MODEL_SCALE to 12.
That makes me realise why we may want the vm to compile with different sizes.
With scale at 12, Cos/Sin currently fail which is expected. I'm reverting scale to 8 in that pull request, acknowledging the gpt2 test would otherwise pass.
Next Pull request is exactly tackling the scale issue for Cos/Sin. I'll also make sure that softmax can run for all relevant scales.

Forpee:
@AntoineF4C5 sounds good!

PR #252, opened 2026-06-17
perf: prove Add/Sub/Neg without a sumcheck (non-ZK path)

### LINKED ISSUE #174: Remove unnecessary sumcheck for Add/Sub/Neg ops
### Context
Add/Sub/Neg all have the same idea that for a tensor with n elements, `∀x ∈ {0, 1}ˡᵒᵍ ⁿ, Add(x) = In0(x) + In1(x)`.
This equation represents an equality of multilinear polynomials, hence to prove it, with committed In0 and In1, we only need to randomly chose an evaluation point `r ∈ Fˡᵒᵍ ⁿ` and assert that `Add(r) = In0(r) + In1(r)`.

### Current proving flow
We currently prove an Add/Sub/Neg by performing a sumcheck: `Add(r) = ∑ eq(r, x) · ( In0(x) + In1(x) )`
This is a sumcheck over n elements, with log n rounds which can be discarded using the above.

### Proposition
The proving flow should just be to recover the evaluation point of the output of the Add/Sub/Neg, and evaluate the inputs at this same point to finally verify the equation holds.

Praz314159:
Closes #174.

`Add`, `Sub`, and `Neg` are linear element-wise ops. The output MLE is a fixed linear combination of the operand MLEs over the same boolean hypercube. Hence `output ≡ Σ±operand` as multilinear polynomials, which by Schwartz–Zippel holds at the already-random output opening point `r`. The per-op sumcheck is therefore redundant.

### Change
These ops now use the no-sumcheck opening-propagation form already used by `Identity`: the prover reads the node's reduced output opening, evaluates the operands at that point `r`, appends them as input openings, and emits **no** execution proof; the verifier checks the linear relation directly (`left+right` / `left-right` / `-operand == output`). Operand openings are discharged by the existing batched HyperKZG opening.

### Scope
The ZK (BlindFold) path still proves these ops via its batched sumcheck — there the algebraic relation is enforced by R1CS constraints derived from the sumcheck `Params`, and the no-sumcheck path has no equivalent constraint yet. So `*Params`/`*Prover`/`*Verifier` are **retained** for the ZK path and `zk.rs` is untouched. A follow-up can add a BlindFold algebraic-identity constraint (cf. the `AlgebraicIdentity` note in `zk.rs`) to drop the sumcheck there too.

### nanoGPT Measurements
Summed prove-span time across the model's 35 `Add`/`Sub` ops:

| | op-prove time |
| --- | --- |
| main (sumcheck) | 44.6 ms |
| this PR (no-sumcheck) | 7.75 ms |

~5.8× faster on those ops (~37 ms off prove). It's a small fraction of nanoGPT's ~2.9 s prove, but scales with residual/LayerNorm-heavy models.

### Soundness
Instead of forging operand claims after a sumcheck, sub soundness test forges the left operand opening directly, so the verifier's direct `left - right == output` check rejects it **at the Sub node** (a tighter rejection than the previous downstream Const-claim mismatch).

### Verification
- `cargo fmt --check`; clippy clean on both `--workspace --all-targets` and
  `-p jolt-atlas-core --features zk --all-targets`.
- `test_add`/`test_sub`/`test_neg` (incl. non-power-of-two) pass.
- Full `jolt-atlas-core` suite (incl. nanoGPT/minigpt/transformer e2e and proof
  serialization) passes.
- `soundness_tests` pass (both Sub `#[should_panic]` tests still reject).
- `--features zk` suite passes (ZK path unchanged).

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/ops/eval_reduction.rs
+++ b/jolt-atlas-core/src/onnx_proof/ops/eval_reduction.rs
@@ -287,10 +287,6 @@ mod tests {
             &mut eval_reduction_proofs,
         );
 
-        // `Sub` is a no-sumcheck op (it only propagates operand openings), so a
-        // Sub-only model emits no execution proof. The eval-reduction proof for
-        // the output node is the artifact that confirms `iop` ran in isolation.
-        assert!(proofs.is_empty());
         assert!(eval_reduction_proofs.contains_key(&output_idx));
     }
 
--- a/jolt-atlas-core/src/onnx_proof/ops/eval_reduction.rs
+++ b/jolt-atlas-core/src/onnx_proof/ops/eval_reduction.rs
@@ -287,7 +287,10 @@ mod tests {
             &mut eval_reduction_proofs,
         );
 
-        assert!(!proofs.is_empty());
+        // `Sub` is a no-sumcheck op (it only propagates operand openings), so a
+        // Sub-only model emits no execution proof. The eval-reduction proof for
+        // the output node is the artifact that confirms `iop` ran in isolation.
+        assert!(proofs.is_empty());
         assert!(eval_reduction_proofs.contains_key(&output_idx));
     }
 
--- a/jolt-atlas-core/src/onnx_proof/malicious_prover.rs
+++ b/jolt-atlas-core/src/onnx_proof/malicious_prover.rs
@@ -10,17 +10,12 @@ use atlas_onnx_tracer::{
 };
 use joltworks::{
     field::JoltField,
-    poly::{
-        commitment::commitment_scheme::CommitmentScheme,
-        opening_proof::ProverOpeningAccumulator,
-        unipoly::{CompressedUniPoly, UniPoly},
-    },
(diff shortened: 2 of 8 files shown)


jolt-atlas-core/src/onnx_proof/ops/eval_reduction.rs
```diff
@@ -287,7 +287,10 @@ mod tests {
             &mut eval_reduction_proofs,
         );
 
-        assert!(!proofs.is_empty());
+        // `Sub` is a no-sumcheck op (it only propagates operand openings), so a
+        // Sub-only model emits no execution proof. The eval-reduction proof for
+        // the output node is the artifact that confirms `iop` ran in isolation.
+        assert!(proofs.is_empty());
```
AntoineF4C5: We should be able to remove this assertion, since it does not really have to do with the testing of correct eval_reduction.


```suggestion
```

AntoineF4C5 (approved):
Looks good to me!
Thank you for applying this change, let's create a follow-up issue to apply to the zk protocol.

PR #254, opened 2026-06-23
Add initial CLI release binaries

### LINKED ISSUE #172: Feature request: CLI installation
It would be great to have an easy-to-use CLI tool (e.g. a `jolt-atlas` or `atlas` binary) for generating and verifying proofs — similar to what ezkl, mina zkml, or other zkml frameworks provide.

alogotron:
## Summary
- add an initial `jolt-atlas` binary with `prove <model>` for bundled fixture models
- add a release workflow that builds tagged CLI archives for Linux, macOS, and Windows
- add a small `scripts/install.sh` helper for downloading release assets

## Notes
This is a conservative first CLI slice for #172. The current repository exposes model-specific proof examples rather than a generic input format, so the binary wraps existing bundled fixture flows instead of inventing a broader ONNX/input CLI contract.

## Verification
- `cargo fmt --all`
- `cargo check --bin jolt-atlas`
- `sh -n scripts/install.sh`
- parsed `.github/workflows/release.yml` with PyYAML

Closes #172.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,68 @@
+name: Release CLI binaries
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - "v*"
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build ${{ matrix.asset }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            asset: jolt-atlas-linux-x86_64
+          - os: macos-14
+            target: aarch64-apple-darwin
+            asset: jolt-atlas-macos-aarch64
+          - os: macos-13
+            target: x86_64-apple-darwin
+            asset: jolt-atlas-macos-x86_64
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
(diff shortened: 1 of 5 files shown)


alogotron:
Thanks for the review! Addressed all points in `e9f8ee4`:

- **Installed binary couldn't find fixtures (`main.rs`, `release.yml`):** The release archive now bundles each fixture under `models/<name>/network.onnx`, and the CLI resolves them via `JOLT_ATLAS_MODELS_DIR` → an exe-relative `models/` directory → the repo path as a dev fallback. `jolt-atlas prove <model>` now works from a plain install.
- **`Model::load` panic (`main.rs`):** The model file is now existence-checked first and surfaced as a structured error instead of a panic.
- **Windows `.exe` handling (`install.sh`):** The extracted/installed filename is tracked separately as `BIN_FILE` (`jolt-atlas.exe` on Windows).
- **Unsupported `os/arch` 404 (`install.sh`):** Added an explicit guard listing the prebuilt targets and a `cargo install` fallback, so unsupported combinations fail with a clear message rather than a raw curl 404.
- The installer also copies the bundled `models/` next to the binary when present.

Local checks: `cargo fmt --all`, `cargo check --bin jolt-atlas`, `sh -n scripts/install.sh`, and YAML parse all pass.

PR #255, opened 2026-06-23
split ps-shout for binary and unary ops

Forpee:


joltworks/src/lookup_tables/prefixes/lower_word_no_msb.rs
```diff
@@ -64,20 +58,10 @@ impl<const XLEN: usize, F: JoltField> SparseDensePrefix<F> for LowerWordNoMsbPre
...
-        let jj = j - XLEN;
-        let x_shift = XLEN - jj;
-        let y_shift = x_shift - 1;
-        let r_x = if j == (XLEN + 1) {
-            F::zero()
-        } else {
-            r_x.into()
-        };
+        let x_shift = XLEN.saturating_sub(j);
+        let y_shift = x_shift.saturating_sub(1);
```
AntoineF4C5: Why using saturating substraction here ? 
I believe j should never exceed `XLEN-1`, using saturating sub suggests the opposite.


```suggestion
        let x_shift = XLEN - j;
        let y_shift = x_shift - 1;
```

joltworks/src/subprotocols/ps_shout/mod.rs
```diff
@@ -0,0 +1,642 @@
...
+
+/// Common interface shared by both binary and unary ps_shout providers.
+///
+/// Contains the three methods that are structurally identical across
+/// `binary::PrefixSuffixShoutProvider` and `unary::PrefixSuffixShoutProvider`,
+/// so implementations only write them once regardless of which variant is used.
+pub trait RafShoutProvider<F: JoltField, LUT: JoltLookupTable + Default> {
+    fn ra_poly(&self) -> (VirtualPoly, SumcheckId);
+    fn r_cycle(&self, accumulator: &dyn OpeningAccumulator<F>) -> OpeningPoint<BIG_ENDIAN, F>;
+    fn table(&self) -> LUT {
```
AntoineF4C5: I believe it is best to remove usage of default methods for traits if none of the implementations override it. Here we can replace any call to this method with `LUT::default`.

AntoineF4C5 (approved):
Nice work!
I made a few comments, but overall great!

joltworks/src/lookup_tables/prefixes/lower_word_no_msb.rs
Forpee: indeed `j` should never exceed `XLEN-1` when using the unary ps-shout. However in the binary case since we maintain and update _all_ prefix checkpoints. `j` does exceed it in that case (as `j` goes up to `64` while `XLEN` is still 32). I will add a `HACK` comment explaining why we use saturating sub here

PR #256, opened 2026-06-29
feat: Add saturation to prover

### LINKED ISSUE #133: Assert soundness in case of a long chain of operators
The current workflow derives from a16z's jolt in a number of way, and one of them is that we do not get the output from a lookup table for each cycle. 
We could argue that in jolt addition, subtraction and multiplication are performed directly on the field elements rather than through a lookup. However, where this is indeed the case, the output of this add/sub/mul is still sent into a RangeChecking lookup table.
This is indeed important, since the output of the addition of two 32-bits field elements is a 33-bits field element, and their product is a 64-bits element, we need this rangecheck lookup table to ensure the cycle output lies in 32-bits.

An important thing to note is that performing add/sub/mul on numbers bigger than u32 doesn't break the laws of those operations, when considering the 32 lowest bits of the result. This means that we can use big numbers to perform arithmetic on them, as long as their 32 lowest bits correspond to those of the expected inputs/outputs.

I currently found two situations where using big numbers could still cause problems:
## Using lookup-table dependant operators:
In this situation, we build a one-hot encoding RA of the input, and prove it correctly represents the input. We are interested in the Raf proving, which asserts that the index of the 1 in the one-hot encoding corresponds to the input value, i.e. RA(input) = 1
To do so, we use an Identity polynomial, which returns I(x) = x, and compute the dot-product of this polynomial's evaluations vector with RA's evaluations.
Since RA is zero except at `index`, the result equals RA(index) * I(index) = 1 * index.
The Raf sumcheck asserts equality between the input and this dot-product, and as the potential values of the input get bigger, the Identity Polynomial takes more variables and the sumcheck needs more rounds and memory.
Jolt uses 64-rounds sumcheck partly because it uses 2 32-bits inputs, but also because when multiplying, the output holds in 64 bits.
In this case, using big numbers (higher than 2^32) would either break the raf-proof, or require really high number of rounds.

#### Mitigation : 
Prepend a "modulo" node, that returns x % 2^32, to ensure the lookup inputs are in the range.

## Field order overflow:
In this situation, we chain enough mul nodes so that the output wraps around the field order.
Since each multiplication doubles the number of bits of its inputs, when starting with 32-bits number, we can get to this point in as low as 3 steps for an order of 2^256 with a binary tree of multiplications: 32bits -> 64bits -> 128bits -> 256 bits.
The output of this last node could be mistaken as a 32bits number, but the most important is that its 32 lowest bits are different from what they should be, now making any arithmetic wrong.
As an example, lets take 3-bits number and a field order of 67.

First notice arithmetics work on big numbers as long as we consider its 3 lowest bits.

let x (y) represent a number and (its restriction to 3 bits)
Hence we can write 9 (1) since 9 % 2^3 = 1
This means that we can use 9 in our vm to represent 1 (9 could appear when we add 7 to 2, which on 3bits equals 1)

```text
Addition:
10 (2) + 12 (4) =  22 (6) | Indeed 22 % 8 = 6
Multiplication:
11 (3) * 10 (2) = 110 (6) | Indeed 110 % 8 = 6
```

Now let's compute this on a field with order 67.
```text
11 (3) * 10 (2) = 110 % 67 = 43 (3) =! 6
```
We can see that once wrapping occurs, we can no longer rely on the arithmetics to old even on the lowest bits.

#### Mitigation :
Use lower inputs (for example 10bits) and making sure the number and chaining of the nodes do not allow for an overflow of the field order. For example, 10-bits inputs and using only additions allow to use up to ~240 nodes.
We can also use the fact that the output of a lookup has known size, so we could use "checkpoint" lookup tables that recover 32-bits numbers. (note the first problem with raf-proving of lookup tables).

### LINKED ISSUE #190: Prove i64-accumulate + fused_rebase + clamp for einsum, mul, cube, etc.
In fixed-point quantization at scale s, a value `x` is represented as `x_hat = x · 2^s`. After a multiplication `x_hat *  y_hat, the result is at scale `2s` and must be "rebased" back to scale `s` by dividing by `2^s`.

**The old path** did this in two separate steps:

1. Multiply in `i32`: `a_i32 * b_i32 -> product_i32`
2. Rebase in a follow-up node: `product_i32 >> scale`

The problem is step 1. Two scaled i32 values multiplied together easily exceed the i32 range (±2^31). For example at scale 12, two modest values like `x_hat = 50,000` and `y_hat = 50,000` produce `2.5 × 10^9` - already near `i32::MAX`. In an einsum (matrix multiply) you're *summing* many such products, so overflows are essentially guaranteed. In Rust release builds this wraps silently, producing garbage values with flipped signs.

The subsequent rebase divides by 2^12, but it's dividing *already-corrupted* wrapped values — the damage is irreversible. Worse, this corruption compounds through every layer: LayerNorm, attention scores, FFN - each multiplication re-wraps, and each rebase re-divides garbage.

**The new path** fuses multiply + accumulate + rebase into a single operation:

1. Upcast to `i64`: `a_i32 as i64 * b_i32 as i64 → product_i64`
2. Accumulate in`i64` (for einsum: sum all products in i64)
3. Rebase in `i64`: `accumulated_i64 / (1 << scale)` — the division happens while the full-precision result is still intact
4. Clamp to `i32` range: `result.clamp(i32::MIN, i32::MAX) as i32` — saturates instead of wrapping, so extreme values hit the rail rather than flipping sign

This matters because in `i64` there's `~9.2 × 10^18` of headroom. A typical GPT-2 matrix multiply accumulates ~768 products of two scale-12 values — the worst-case sum is well within`i64` range. So the rebase division now operates on the *exact* accumulated result, and the final clamp ensures no sign-flip corruption leaks into downstream layers.

The effect is multiplicative across layers: fixing the precision at each multiply/einsum means LayerNorm inputs are correct, which means attention scores are correct, which means softmax outputs are correct, etc. That's why the end-to-end metrics jump from 0% top-1 agreement to 100%.

### LINKED ISSUE #193: Refactor Einsum prover/verifiers: deduplicate `bmk_*` and `mbk_*` variants
In the Einsum proof system, a few equation variants appear to share essentially the same proving/verifying structure, with only small differences in how operands are indexed/transposed:

`bmk_bkn_mbn` and `bmk_kbn_mbn`
`mbk_bnk_bmn` and `mbk_nbk_bmn`

Today these are implemented as separate prover/verifier modules and dispatched independently in [mod.rs](https://github.com/ICME-Lab/jolt-atlas/blob/main/jolt-atlas-core/src/onnx_proof/ops/einsum/mod.rs)

# Proposal
Refactor the prover(s) so each “pair” of structurally-similar einsums shares one implementation:
One shared implementation for the `bmk_*bn*->mbn pair`
One shared implementation for the `mbk_*bk->bmn pai``

### LINKED ISSUE #194: implement prover logic for monolithic `MeanOfSquares` op
see here for onnx runtime logic: https://github.com/ICME-Lab/jolt-atlas/blob/main/atlas-onnx-tracer/src/ops/mean_of_squares.rs

### LINKED ISSUE #203: ScalarConstDiv remainder is committed but never range-checked

Forpee:
Part of the ongoing effort for [rqe-milestone](https://github.com/ICME-Lab/jolt-atlas/milestone/7)

- [x] #190 
- [x] #190 

Note: gpt2 test does pass locally, but with the new changes it now takes too long and gets killed on ci. This is primarily due to `PCS::prove` taking over a minute (locally) now that `model.max_num_vars()` has increased from `20` to `24`. We could potentially mitigate this by using dory although that would introduce significant engineering overhead around how we batch our polynomials vs jolt

---
additional work: 
- [ ] update the zk path
- [ ] rm dummy clamped tanh claim I added


joltworks/src/lookup_tables/sat_clamp.rs
AntoineF4C5, line 1: The Saturating lookup table introduces a confusion between the table size and word size.
Globally, the table size is 64 bits for binary tables, 32 for unary table. The word size for both is 32, since both deal with 32 bits words.
The saturating clamp is slightly ambiguous since it's input is an 64 bits (i64), but the output is represented over 32 bits (i32).

Hence, considering XLEN=64 for saturating to i32 is a bit confusing, but is fine. Let's just add a small comment for that matter.

AntoineF4C5 (approved):
Looks good to me ! 
Added a comment concerning the XLEN for Saturating lookup table.

Forpee:
Merging despite CI failures as gpt2 passes locally and the CI failure is due to test running too long. I'll address the long proving time in a follow up.

I'm also deferring zk path updates from this PR due to time constraints. This is tracked in #28 and will be handled in a follow up

PR #262, opened 2026-07-04
fix(tracer): drop the consumed exponent input in `Pow` lowering — un-breaks `Pow(x,3)` proving

### LINKED ISSUE #261: `Pow(x, 3)` models panic at prove: lowered `Cube` keeps the broadcast exponent as a 2nd operand (`cube.rs:143`)
While running measurement sweeps over the in-tree example models I found that the two `nanoGPT_1M` exports cannot be proven on `main` (`8fcd92c`; re-verified 2026-07-05):

```
thread 'main' panicked at jolt-atlas-core/src/onnx_proof/ops/cube.rs:143:13:
Expected one operand for Cube operation
```

**Affected / not affected (measured on `8fcd92c`; prove/inspect ledger available on request):**

| model | cubic in source | lowered graph (inspected) | prove |
|---|---|---|---|
| `models/nanoGPT_1M` (5L) | GELU `x**3` → ONNX `Pow(x,3)` | 5 `Cube` nodes, **each with 2 inputs** `[base, Broadcast(exponent)]` | panic `cube.rs:143` |
| `models/nanoGPT_1M_deep` (20L) | same | 20 `Cube` nodes, same 2-input wiring | panic `cube.rs:143` |
| `models/transformer` | GELU `0.044715 * x * x * x` | no `Cube` and no GELU `Pow` — the coefficient multiplies first, so the cubic lowers to plain binary `Mul` nodes | PASS |
| synthetic `y = x**3` ([1,16], opset 10) | `Pow(x,3)` | 1 `Cube`, 2 inputs `[Input, Broadcast]` | panic `cube.rs:143` |
| synthetic `y = x**2` | `Pow(x,2)` | 1 `Square`, **1 input** (exponent gone) | PASS |
| synthetic `y = x*x*x` | `Mul(Mul(x,x),x)` | `Square` (1 input; the `Mul(x,x)` pair is folded) + binary `Mul` | PASS |

**Root cause (verified in the pinned sources):** two pieces interact.

1. tract's `declutter_pow` (pinned `sonos/tract` @ `c484b3ee`, `core/src/ops/math/mod.rs:476`) rewrites `Pow(x, 2.0)` to the unary `square()` op wired to **only the base input** (and `Pow(x, 0.5)` to `sqrt()`), but has **no rule for exponent 3** — so `Pow(x,3)` reaches atlas-onnx-tracer's `handle_pow` still carrying its exponent input, while `Pow(x,2)` never does.
2. `handle_pow` → `build_cube` (`atlas-onnx-tracer/src/node/handlers/arith.rs`) builds the `Cube` node via `HandlerBuilder::simple_op`, which clones **all** of the handler's `internal_input_indices` onto the node — so the (broadcast) exponent constant survives as a second graph input. `Trace::layer_data` then returns **two** operands, and `CubeProver::initialize` requires exactly one:

```rust
let [operand] = operands[..] else {
    panic!("Expected one operand for Cube operation")
};
```

`square.rs` has the identical destructure, but it never trips: every LayerNorm variance (`Pow(x,2)`) in these same graphs lowers to the 1-input `Square` form (operand count 1 measured on all 11 / 41 `Square` nodes of the two models), and `Pow(x,2)`-only graphs prove end-to-end (the `transformer` model and the `y = x**2` synthetic above). Tracer-level execution is unaffected for all of the above (trace generation and witness decoding work; the panic is prove-only).

One more data point that narrows it down: the GPT-2 example also carries `Pow(x,3)` (12 tanh-GELU cubics, measured on the CI export) yet `test_gpt2` passes — because it runs with `pre_rebase_nonlinear = true`, whose decomposed path (`ScalarConstDiv` → `Mul` → `Mul`) uses only the base input and never builds a `Cube` node. The panic is specific to the **default** lowering path.

**Minimal repro** (the export settings mirror the in-tree `gen.py` files):

```python
import torch
class Pow3(torch.nn.Module):
    def forward(self, x):
        return x ** 3
x = 0.1 * torch.rand(1, 16)
torch.onnx.export(Pow3().eval(), x, "network.onnx", export_params=True,
                  opset_version=10, do_constant_folding=True, dynamo=False,
                  input_names=["input"], output_names=["output"],
                  dynamic_axes={"input": {0: "batch_size"}, "output": {0: "batch_size"}})
# on torch builds without the `dynamo` kwarg, drop it (the legacy exporter is then the default)
# replacing `x ** 3` with `x * x * x` (or `x ** 2`) proves fine
```

then prove with the standard public API (same calls as `e2e_tests.rs::prove_and_verify`; any input triggers it — the failure is structural):

```rust
let model = Model::load("network.onnx", &Default::default());
let input = Tensor::new(Some(&data), &[1, 16]).unwrap(); // e.g. quantized ~1.0 values
let pp = AtlasSharedPreprocessing::preprocess(model);
let prover = AtlasProverPreprocessing::<Fr, HyperKZG<Bn254>>::new(pp);
let _ = ONNXProof::<Fr, Blake2bTranscript, HyperKZG<Bn254>>::prove(&prover, &[input]);
// -> panicked at cube.rs:143: Expected one operand for Cube operation
```

For the in-tree models: load `models/nanoGPT_1M/network.onnx` with its `input.json` tokens (`[1,128]`, vocab 65) through the same API — same panic.

**Why CI never caught it:** there is no e2e test for either `nanoGPT_1M` export, and no other e2e-covered model contains a `Pow(x,3)` node (`transformer` writes its GELU cubic as `x * x * x`). The op-level `test_cube_zk` doesn't cover this either — it wires a 1-operand `Cube` directly via `ModelBuilder`, so the `Pow` lowering path is never exercised.

**Possible fixes:**
1. tracer-side — have `handle_pow`/`build_cube` wire only the base input on the `Cube` node (dropping the exponent, exactly what tract's exponent-2 declutter already produces for `Square`); or
2. prover-side — have `CubeProver::initialize` take `operands[0]` explicitly.

Option 1 seems cleaner (keeps the prover contract "one operand" intact and makes `Cube` consistent with `Square`) — **I'm sending a small PR implementing it**, with a tiny `Pow(x,3)` fixture pinning the path; flagging option 2 in case you'd rather take the prover-side route. Possibly related to the ongoing Cube work tracked in #190.

Environment: `main` @ `8fcd92c`, repo-pinned toolchain (rustc 1.88), macOS arm64; exports made with torch 2.12.1 (opset 10, legacy exporter).

pdezk:
Fixes #261.

**Problem.** Graphs containing `Pow(x, 3)` cannot be proven: the lowered `Cube` node keeps the (broadcast) exponent constant as a second input on the node, so `Trace::layer_data` hands `CubeProver::initialize` two operands and its `let [operand] = operands[..]` destructure panics (`cube.rs:143`, "Expected one operand for Cube operation"). This breaks both in-tree `nanoGPT_1M` exports (their GELU is written `x**3`; 5 and 20 `Cube` nodes respectively), which have never been provable with the default run args.

Why only exponent 3: tract's `declutter_pow` rewrites `Pow(x, 2)` to its unary `Square` op (exponent dropped) before the handlers run, but has no rule for exponent 3 — so `Pow(x,3)` is the one power that reaches `handle_pow` still carrying its exponent input, and `build_cube`'s `simple_op` clones all handler inputs onto the node. (The GPT-2 example also contains 12 `Pow(x,3)` GELU cubics but escapes via `pre_rebase_nonlinear = true`, whose decomposed path uses only the base input — the panic is specific to the default lowering path.)

**Fix.** Drop the exponent from the handler inputs in `handle_pow` once it has been consumed by the dispatch — the lowered `Square`/`Cube` node becomes unary, matching what the tract declutter already produces for exponent 2. `HandlerContext` is constructed per node, so the truncation is local to this one lowering. Before truncating, the exponent is now asserted to be a uniform integer constant, so unsupported `Pow` graphs (fractional or per-element exponents) still fail loudly — at lowering time with a clear message, instead of the old prove-time operand panic (or, post-fix, silently proving a floored exponent). Also covers the latent `build_square` instance of the same wiring; no prover-side change — the "exactly one operand" contract stays intact.

**Test.** Adds a 222-byte `models/pow_cube` fixture (`y = x**3`, opset 10 — the ONNX is `[Constant, Pow]`; note `x*x*x` or `x**2` would not cover this path) and `test_pow_cube`, which pins both the panic regression and, via an output-variability assertion, that the operand surviving the lowering is the base and not the broadcast exponent.

**Validation** (full run logs available on request).

| check | result |
|---|---|
| `test_pow_cube` (new) | PASS (0.03 s) — negative control: without the fix it panics at `cube.rs:143` |
| `models/nanoGPT_1M` prove+verify | **PASS, 6.3 s** (first proof of this export on the default path) |
| `models/nanoGPT_1M_deep` prove+verify | **PASS, 10.3 s** (same) |
| lowered graph re-inspect | every `Cube` now 1-input; the dangling exponent `Constant`+`Broadcast` pair is pruned (nanoGPT_1M graph: 473 → 463 nodes) |
| `cargo fmt --all --check` / `clippy` default + `zk` (`RUSTFLAGS=-D warnings`) | clean |
| `cargo nextest run --workspace` / `-p jolt-atlas-core --features zk` | 296/296 and 132/132 pass |
| `test_gpt2`, `test_bge_small_en_v1_5` (ignored-only jobs) | PASS (17.4 s / 8.8 s) — `test_gpt2` exercises the `pre_rebase_nonlinear` path with 12 real `Pow(x,3)` nodes |
| wasm build (`atlas-onnx-tracer`, `wasm32-unknown-unknown`) | builds |

If you'd rather pin the two `nanoGPT_1M` exports directly, I'm happy to add e2e tests for them as well (~6.5 s / ~10 s on my machine) — the tiny fixture keeps the default CI cost negligible.


Forpee:
Hi @pdezk thank you for noticing this bug and implementing a fix. I believe this will be resolved when #256 lands. Please comment there if there is anything else you see regarding this.

pdezk:
Thanks @Forpee — confirmed, #256 lands the fix, so I'm closing this as redundant.

Re-verified on `main` @80a718e (the #256 merge): `handle_pow` now truncates `internal_input_indices` to 1, so the lowered `Cube` node is unary and the broadcast exponent no longer reaches the prover as a second operand. Measured prove+verify:

| model | pre-#256 (8fcd92c) | main @80a718e |
|---|---|---|
| `nanoGPT_1M` | panic `cube.rs:143` | ✅ prove+verify |
| `nanoGPT_1M_deep` | panic `cube.rs:143` | ✅ prove+verify |
| minimal `Pow(x,3)` fixture | panic `cube.rs:143` | ✅ prove+verify |

Thanks again!

PR #266, opened 2026-07-07
fix(zk): bring the zk path up to date with the non-zk prover (#28 regression from #256)

pdezk:
## Problem

#256 restructured the non-zk prover (i64 accumulate + saturating clamp for `Add`/`Sub`/`Sum`, fused
rescale+clamp for einsum/`Mul`/`Square`/`Cube`, a clamped-input advice for `Tanh`) and deferred the
mirroring zk updates to #28. Since the merge, the `Clippy & Test (zk)` job is red. Concretely, on
`main` the zk path has two breaks:

1. **Compile** — `TanhProver::initialize` gained `(acc, transcript)` parameters (it now appends the
   `VirtualPoly::DummyClampedTanhInput` advice), but the shared `prove_lookup_zk!` macro in
   `onnx_proof/zk.rs` still calls the 2-arg form → `E0061`. This is the only compile break
   (audited statically and re-confirmed empirically by stubbing the Tanh arm and rebuilding all
   targets under `-D warnings`).
2. **Prove-time panic for fused-rescale ops** — #256 folded the rescaling division and saturating
   clamp *into* einsum/`Mul`/`Square`/`Cube` (the tracer no longer emits the separate
   `ScalarConstDiv` rebase node). Their arithmetic sumcheck's input claim is now
   `ClampAcc·2^S + RescaleRemainder` — two advices appended by `fused_rebase::prove_pre` /
   `clamp_lookups`, which the zk pipeline never runs. Any fused node panics at prove time with
   `opening for ... ClampAcc(i) not found`. With the Tanh arm unblocked, 10 of 150 zk tests fail —
   nine at exactly this panic: `test_{square,mul,cube,einsum}_zk`,
   `test_multi_op_{arithmetic_chain,mul_div_sub}_zk`, `test_srs_derived_pedersen_generators`,
   `test_zk_rejects_rebound_input_for_square_model`, and
   `ops::square::tests::test_square_blindfold_e2e`. The tenth, `test_multi_op_cube_relu_zk`,
   trips a stale fixture that the same #256 fusion change exposed one step earlier, at trace
   time (measured: `attempt to shift left with overflow` in the tracer's `floor_rebase`,
   `ops/mod.rs:229` — see Tests).

## Change

- **Tanh (compile + advice threading).** `prove_lookup_zk!` accepts optional trailing args
  forwarded to `initialize`; the Tanh arm passes `&mut prover.accumulator, &mut prover.transcript`
  (Sigmoid/Erf stay 2-arg). The advice append is transcript-quiet in zk mode on both sides
  (`TanhVerifier::new` already appends the verifier-side placeholder), so prover/verifier
  transcripts stay in sync, and the claim is queued for BlindFold exactly like other zk-mode
  advice claims.
- **Fused-rescale ops (einsum/`Mul`/`Square`/`Cube`).** New `prove_fused_rebase_{pre,post}_zk` /
  `verify_fused_rebase_{pre,post}_zk` in `onnx_proof/zk.rs` mirror the non-zk flow around the
  (unchanged) arithmetic sumcheck:
  - *pre*: append the `RescaleRemainder` advice, then the 64-bit saturating-clamp read-raf lookup
    (`output = SatClamp(rescaled)`, appends `ClampAcc`) and its one-hot batch over `ClampRaD` —
    the same `ps_shout` read-raf + `shout::ra_onehot` building blocks the zk path already uses for
    ReLU and the neural-teleport range checks, run through the same
    `run_zk_sumcheck`/`run_zk_batched_sumcheck` machinery;
  - *post*: the rescaling-remainder identity range check `R ∈ [0, 2^S)` and its one-hot batch over
    `RescaleRemainderRaD` (the identity-RC params already carry BlindFold constraint impls).
  The verifier consumes the same four extra stages per fused node, in order.
- **Commit set.** The #256 `ClampRaD` commit filter becomes conditional: fused-rescale nodes now
  commit *and open* `ClampRaD` (and their `RescaleRemainderRaD`); `Add`/`Sub`/`Sum` keep the
  filter because their zk provers still prove the un-clamped relation (see "Not in this PR").
- **Scalar fused nodes** (`log_T = 0`): explicit `unimplemented!` on both sides. The non-zk scalar
  fallback checks the clamp on cleartext claims, which has no zk counterpart yet; failing loudly
  beats silently proving an unbound clamp. (Previously these panicked with the opaque `ClampAcc`
  lookup error.)
- **Tests.**
  - Six zk tests bump `PedersenGenerators::deterministic(16)` → `32`: the clamp read-raf stage
    raises the Hyrax column count to 32 — the size every lookup-op zk test (sigmoid/tanh/erf/relu/
    einsum/sum) already uses. The existing generator-size assert caught this cleanly.
  - `test_multi_op_cube_relu_zk`: `cube(i, 1 << 8)` → `cube(i, 1)`. Pre-#256 the fused Cube kernel
    sat behind the default-off `fused-ops` feature, so `scale` was ignored (raw `x³`); #256 made
    the kernel unconditional, turning `scale = 256` into a 2·256 = 512-bit shift that overflows at
    trace time. Stale fixture (the non-zk path panics identically on this model; #256's own cube
    test uses `cube(i, 12)`).
  - `ops::square::tests::test_square_blindfold_e2e` drives `SquareProver` directly (bypassing the
    zk.rs orchestration), so it now appends the two advices itself, exactly as
    `prove_fused_rebase_pre_zk` does.

## Not in this PR (documented, deliberate)

- **`Add`/`Sub`/`Sum` still prove the un-clamped relation in zk** — #256's explicit deferral (the
  in-code comments in `zk.rs` and `ops/add.rs` say the clamp fold into the zk path is tracked
  separately). Their zk tests pass on non-saturating traces; wiring the clamp lookup for them is
  the remaining #28 body of work and follows the exact stage pattern this PR establishes for the
  fused ops (happy to do that as a follow-up).
- **`MeanOfSquares` zk prove/verify `unimplemented!`** — pre-existing gap, predates #256. (This is
  also what `test_gpt2_zk` — `#[ignore]`, not a CI job — stops on; on `main` that test cannot run
  at all since the zk build doesn't compile.)
- **Scalar fused-rescale nodes in zk** — loud `unimplemented!` (above), not silently weakened;
  mirrors the `MeanOfSquares` arm's convention.
- Two pre-existing observations surfaced while auditing (no code change here): `fused_input_claim`'s
  `1u64 << bits` wraps silently in release for `bits ≥ 64` (absurd-but-expressible scales; the stale
  `cube(i, 1 << 8)` fixture sat in exactly that regime — its debug build panics first in the
  tracer's `floor_rebase`), and the verifier-side `unimplemented!` arms panic
  rather than returning `Err` on hostile bundles.

## Soundness

Parity with the shipped zk design — no new trust assumptions, and none removed:

- The fused-op stages reuse, unmodified, the constraint-carrying sumcheck types that already run
  under BlindFold elsewhere in the zk path (ReLU's `ps_shout` read-raf, the teleport one-hot
  batches, the identity range check). Each new stage's *output* claim is therefore bound by the
  same BlindFold constraint set as its non-zk counterpart's verifier checks: the read-raf output
  relation, one-hot/booleanity/hamming over the committed `ClampRaD`/`RescaleRemainderRaD`, and
  the identity-RC output relation.
- Stage *initial* claims — including the fused arithmetic stage's
  `ClampAcc·2^S + RescaleRemainder` and the Tanh lookup stage's `rv + γ·DummyClampedTanhInput` —
  are prover-supplied baked constants in the BlindFold R1CS, like **every** chain-start initial
  claim in the zk pipeline today (no non-default `InputClaimConstraint` exists in the codebase;
  Sigmoid/Erf's lookup stages work the same way). The new code's comments state this explicitly
  rather than overclaiming. Wiring real input-claim constraints over the advice openings is a
  natural follow-up and would slot into the existing `StageConfig::with_input_constraint`
  machinery — happy to take that on, but it is new protocol work beyond restoring #256 parity,
  and it belongs with the other baked-claim bindings tracked as future work.
- An adversarial review pass over the change (independent reviewers on Tanh advice flow, stage
  mechanics/transcript sync, zk-vs-non-zk parity, and the test fixtures) found no completeness or
  transcript-sync defects; its substantive findings were the comment-accuracy point above and one
  missed generator bump in an ignored benchmark (both addressed).

## Verification

All on this branch, macOS (Apple Silicon), `RUSTFLAGS='-D warnings'` for the clippy rows:

| Check (CI mirror) | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --workspace --all-targets` | clean |
| `cargo clippy -p jolt-atlas-core --all-targets --features zk` | clean |
| `cargo nextest run --workspace` (default) | **318/318 passed** |
| `cargo nextest run -p jolt-atlas-core --features zk` | **150/150 passed** (base `80a718e`: compile error; Tanh-fix-only: 140 pass / 10 fail) |
| GPT-2 job (`--run-ignored ignored-only -E 'test(test_gpt2)'`, default) | **PASS** (122.7s locally) |
| BGE job | **PASS** (44.3s) |
| wasm build (`atlas-onnx-tracer`, `wasm32-unknown-unknown`) | clean |
| `test_gpt2_zk` (zk, `#[ignore]`; not a CI job) | runs to the **pre-existing** `MeanOfSquares` zk `unimplemented!` (same arm exists verbatim pre-#256; on `main` the zk build doesn't compile, so this test can't run at all) |

## Context

Saw the zk updates were deferred to #28 due to time — took a full pass over the regression surface
(compile + every failing zk test root-caused). Happy to adjust, split, or defer any part if you're
already mid-way on #28.

### DIFF
--- a/jolt-atlas-core/src/onnx_proof/e2e_tests.rs
+++ b/jolt-atlas-core/src/onnx_proof/e2e_tests.rs
@@ -806,7 +806,7 @@ fn test_square_zk() {
 
     let gens = joltworks::poly::commitment::pedersen::PedersenGenerators::<
         joltworks::curve::Bn254Curve,
-    >::deterministic(16);
+    >::deterministic(32);
     let (bundle, io) = crate::onnx_proof::zk::prove_zk(&prover_pp, &[input], &gens);
 
     crate::onnx_proof::zk::verify_zk(&bundle, &verifier_pp, &io, &gens)
@@ -1008,7 +1008,7 @@ fn test_zk_rejects_rebound_input_for_square_model() {
 
     let gens = joltworks::poly::commitment::pedersen::PedersenGenerators::<
         joltworks::curve::Bn254Curve,
-    >::deterministic(16);
+    >::deterministic(32);
 
     // Honest proof for input x.
     let (bundle, mut io) =
@@ -1185,7 +1185,7 @@ fn test_mul_zk() {
     let verifier_pp = AtlasVerifierPreprocessing::<Fr, HyperKZG<Bn254>>::from(&prover_pp);
     let gens = joltworks::poly::commitment::pedersen::PedersenGenerators::<
         joltworks::curve::Bn254Curve,
-    >::deterministic(16);
+    >::deterministic(32);
     let (bundle, io) = crate::onnx_proof::zk::prove_zk(&prover_pp, &[input], &gens);
     crate::onnx_proof::zk::verify_zk(&bundle, &verifier_pp, &io, &gens)
         .expect("ZK verification should succeed");
@@ -1208,7 +1208,7 @@ fn test_cube_zk() {
     let verifier_pp = AtlasVerifierPreprocessing::<Fr, HyperKZG<Bn254>>::from(&prover_pp);
     let gens = joltworks::poly::commitment::pedersen::PedersenGenerators::<
         joltworks::curve::Bn254Curve,
-    >::deterministic(16);
+    >::deterministic(32);
     let (bundle, io) = crate::onnx_proof::zk::prove_zk(&prover_pp, &[input], &gens);
     crate::onnx_proof::zk::verify_zk(&bundle, &verifier_pp, &io, &gens)
         .expect("ZK verification should succeed");
@@ -1688,7 +1688,7 @@ fn test_multi_op_arithmetic_chain_zk() {
     let verifier_pp = AtlasVerifierPreprocessing::<Fr, HyperKZG<Bn254>>::from(&prover_pp);
     let gens = joltworks::poly::commitment::pedersen::PedersenGenerators::<
         joltworks::curve::Bn254Curve,
-    >::deterministic(16);
+    >::deterministic(32);
     let (bundle, io) = crate::onnx_proof::zk::prove_zk(&prover_pp, &[input], &gens);
     crate::onnx_proof::zk::verify_zk(&bundle, &verifier_pp, &io, &gens)
         .expect("Multi-op arithmetic chain ZK verification should succeed");
@@ -1719,7 +1719,7 @@ fn test_multi_op_mul_div_sub_zk() {
     let verifier_pp = AtlasVerifierPreprocessing::<Fr, HyperKZG<Bn254>>::from(&prover_pp);
     let gens = joltworks::poly::commitment::pedersen::PedersenGenerators::<
         joltworks::curve::Bn254Curve,
-    >::deterministic(16);
+    >::deterministic(32);
     let (bundle, io) = crate::onnx_proof::zk::prove_zk(&prover_pp, &[input], &gens);
     crate::onnx_proof::zk::verify_zk(&bundle, &verifier_pp, &io, &gens)
         .expect("Multi-op mul/div/sub ZK verification should succeed");
@@ -1736,7 +1736,10 @@ fn test_multi_op_cube_relu_zk() {
     let input = Tensor::<i32>::random_small(&mut rng, &[size]);
     let mut builder = ModelBuilder::new();
     let i = builder.input(vec![size]);
-    let c = builder.cube(i, 1 << 8);
+    // `scale` is the log2 rebase shift (bits = 2·scale, see `rebase_bits`);
+    // the pre-#256 `1 << 8` predates the fused kernel, which now interprets
+    // it as a 512-bit shift.
+    let c = builder.cube(i, 1);
     let out = builder.relu(c);
     builder.mark_output(out);
     let model = builder.build();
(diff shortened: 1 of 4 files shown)


pdezk:
CI note on the red GPT-2 job: that's the pre-existing failure on `main`, not this PR — `test_gpt2` is aborted with SIGKILL on the hosted runner (`SIGKILL [316.1s]`), and the same job already fails on the merge base `80a718e` itself, as acknowledged in the #256 merge comment ("gpt2 passes locally ... takes too long and gets killed on CI"). Nothing in this diff executes in that job. The changes are zk-gated code plus zk test fixtures. And the job's exact command passes locally on this branch (~123 s). Meanwhile `Clippy & Test (zk)`, which has been red on `main` since #256, is green here. That's the job this PR exists to fix.

PR #272, opened 2026-07-09
Rsqrt2

### LINKED ISSUE #189: Increase rsqrt precision
The tracer now computes `rsqrt(x) = isqrt(S³ / x)` in i64 using the model's actual scale factor (e.g. S = 1 << 12 = 4096 at scale 12). This gives near-perfect agreement with the f32 reference.

We need to update the prover accordingly

Forpee:
* update prover to accurate rsqrt
* update mos to use the correct count for increasing ppl for padded model

Result: gpt2 at 100% (99.99..%) accuracy now (even with padding)


AntoineF4C5 (commented):
LGTM!

PR #274, opened 2026-07-09
feat: reduce re-execution of einsum_acc_i64

Forpee:
The gpt2 prover ran `einsum_acc_i64` 6× per node (~18s): 1× trace gen, 2× witness gen, 3× IOP. The clamp quotient and rescale remainder both come from one accumulation, but every consumer recomputed it.

Fix: compute once per phase, derive both tensors from it (`*_intermediate_and_remainder` kernels + `RebaseIntermediates` threaded through the prover). Net 6 -> 3 (~9s off gpt2)

Why not cache in the trace (->1×)? ~150MB extra resident for gpt2, growing with `seq_len`. The trace stays a lean i32 record; re-execution retains zero memory.


AntoineF4C5 (commented):
LGTM!

PR #275, opened 2026-07-09
Read-Raf refactor, new PrefixSuffix tables tests

AntoineF4C5:
This pull requests sets the ground for the new clamping operator.
The goal of the refactor is to start making the OpLookup generic, and RangeChecking helpers more similar to OpLookup helpers, in a long term goal to have a generic implementation with several stages of specialization for range-check, operator lookups and any other lookup (clamping, saturating ...) used between operators.

Also adds some tests to ease the engineering of new lookup tables, by pointing more efficiently errors to their cause.

Note: Need to implement the Read-Raf test for binary lookup tables, and make it relevant for saturating. Other than that, this PR is ready for review. I might offload the binary Read-Raf test, since we currenty use no binary lookup table except for rangechecking which is stable.


Forpee (commented):
looks good - very easy to add new prefix/suffix variants now & the extra mle-linearity and weighted-evals checks are a nice strengthening to the suite. Happy to proceed. 

One thing to clarify: read-raf test looks like its moving to the lookup-table sites, while `ps_shout::{unary,binary}` still have their own read-raf tests. Is the intent to move those under the table implementations, or keep both?

AntoineF4C5:
@Forpee In deed, I had not seen that unary/binary implement a read-raf test for all lookup tables!
I wonder which is best between leaving them there, of moving to table's implementations, making it less likely for an implementer to forget adding this test when creating a new table, since it's one less file to modify.

Forpee:
if its a straightforward change lets move it to the table impl sites

Issue #279, opened 2026-08-10
Explore "partially prefix-suffix decomposable" tables.


AntoineF4C5:
## Context

We compute `COS[x % 2π]` (and similarly `SIN`) via a lookup into a custom table indexed by `x` reduced modulo the period.

At scale 14, `2π` is approximated with 17 bits, so `x % 2π` needs up to 17 bits to represent all possible values (`x` itself is 32 bits). Since `COS`/`SIN` tables admit no prefix-suffix decomposition (unlike our regular 32-bit tables), their cost scales directly with table size — so we want to keep them small, currently targeting sizes below 16 bits.

To do this, we downscale `x % 2π` to a smaller scale before the table lookup. Downscaling by `D` bits is a right shift `>> D` rather than a division: division would require committing to 2 advice values, whereas a right shift admits a prefix-suffix decomposition (as used in our other lookup tables), making it cheaper. `D` is chosen per scale, derived from both the scale and the target table size.

**Current approach — 3 sumchecks:**
1. Take `x % 2π` (division sumcheck)
2. Downscale the remainder by `D` bits (read-raf sumcheck)
3. Look up the downscaled value into the `COS` table (small-table read-raf)

Steps 2 and 3 each require their own one-hot encoding (of the remainder, and of the downscaled value respectively) — two one-hot encodings committed in total.

## Idea

Reduce this to a single one-hot encoding.

Right-shifting `x` corresponds to `∑_{j≥D} x_j · 2^(j−D)`. This means the bits of the remainder in range `[D, D + log2(COS.len()) − 1]` *are* the downscaled value — no separate computation is needed to derive it.

We could define a custom table that is "unstructured" (arbitrary values) over bits `[D, D + log2(COS.len()) − 1]`, and has trivial/ignorable behavior over bits `[0, D − 1]`. This would let us index directly with the remainder's existing one-hot encoding, while keeping the unstructured portion of the table limited to `log2(COS.len())` bits.

**Expected benefit:** removes one sumcheck round (step 2) and the need to commit a second one-hot encoding.

**Open question:** we likely still need a sumcheck over the low `[0, D−1]` bits to produce a committed evaluation, but with simplified behavior since those bits are ignored/don't affect the table's structured output — exact behavior TBD.

PR #281, opened 2026-08-12
feat: downscale trig table inputs

### LINKED ISSUE #216: Prover infrastructure is rigidly coupled to a fixed single scale
## Problem

The prover assumes a single globally-fixed quantization scale. Many ops have constants and logic hardwired to a specific scale value rather than being parameterized by it. Changing the scale requires a scavenger hunt across the codebase.

### Ops with scale rigidity

| Component | Current state |
|---|---|
| **Rsqrt** | Hardcoded `Q = 256` (i.e. scale 8) in `rsqrt.rs`. Will not work at any other scale. |
| **Neural teleport tables** (tanh, erf, sigmoid) | `SCALE = 256.0` hardcoded in `neural_teleport/mod.rs`. `LOG_TABLE_SIZE` and `TAU` in `activation.rs` are compile-time constants with no relationship to the active scale. |
| **Softmax** | Does not support odd scales (range-check limitation). |
| **Sin / Cos tables** | Built against the hardcoded `SCALE = 256.0` in the neural teleport module. |

### Root cause

There is no centralized contract between the global scale and the ops that depend on it. Each op manages its own scale handling — some accept it as a parameter, others pull from a hardcoded constant, and a few silently assume a specific value. Nothing enforces consistency or flags breakage when the scale changes.

### Desired state

- **Parameterize scale throughout:** Ops like rsqrt and the neural teleport table builder should derive from the runtime scale, not a `const`.
- **Scale-aware table sizing:** Activation ops should derive `log_table` and `tau` from the scale (and ideally from input-range bounds), similar to how `Gather` dynamically sizes its encoding from `dict_len`.
- **Softmax for all scales:** Remove the even-only restriction.

### LINKED ISSUE #273: Scale-aware Cos/Sin teleportation: brute-force the per-scale `k·2π·2^s` modulus (blocks accurate Qwen/RoPE)
## Summary

The cos/sin periodic-teleport path is hardcoded for the reference scale (2⁸). To run Qwen accurately we need higher scales (target: **scale 14**), which requires the trig teleport constants and lookup-table quantization to be derived per-scale rather than baked in.

Today the signed activations (tanh/erf/sigmoid) are scale-aware, but cos/sin are not — they're guarded to reject any scale ≠ 8 in `atlas-onnx-tracer/src/node/handlers/activation.rs` so they can't silently emit a wrong proof.

## What's hardcoded

- **`FOUR_PI_APPROX = 3217`** (`atlas-onnx-tracer/src/model/mod.rs`) — the `k·2π·2ˢ` period modulus, tuned only for s=8. Threaded through the op execution (`eval_trig`), witness, zk, and the cos/sin proofs (~15 sites).
- **`neural_teleport::SCALE = 256.0`** (`jolt-atlas-core/src/onnx_proof/neural_teleport/mod.rs`) — the LUT output quantization, used by `CosTable`/`SinTable`; must become `2^scale`.
- **`COS_LOG_TABLE_SIZE`** — derived from `FOUR_PI_APPROX`, so becomes scale-dependent.

## Approach

**1. Brute-force the period modulus per scale.** The teleport wraps inputs mod `2π` using the periodicity of cos/sin. Because `round(2π·2ˢ)` has quantization error, we search over multiples `k` and pick the `k` minimizing `|round(k·2π·2ˢ) − k·2π·2ˢ|` (per-wrap error). At s=8 this yields `k=2` → `round(4π·256) = 3217`, keeping accumulated error within ~`1/2ˢ` over ~100 wraps. Generalize this to a function of scale.

**2. Quantize the LUT at the model scale** (`2^scale`), matching the op (which evaluates at `scale_to_multiplier(scale)`) — same fix already applied to tanh/erf/sigmoid.

## Table-size constraint (the hard part)

Prover requires **LUT ≤ 2¹⁶** (larger is too slow). At scale 14 this is the binding constraint:

- A **full period** `2π·2¹⁴ ≈ 102,944` → **2¹⁷**, already over budget.
- Dropping the teleport and using a direct table is a non-starter: RoPE angle = position radians (dim i=0, `inv_freq=1`), so a 2¹⁶ table covers only **~4 tokens**.

To fit 2¹⁶ at scale 14, add **cos/sin symmetry (quadrant) folding** on top of the periodic reduction:

| reduced span | table @ s=14 |
|---|---|
| full `2π` | 2¹⁷ ❌ |
| half `π` | 2¹⁶ (tight) |
| **quarter `π/2`** | **2¹⁵ ✅** |

Quarter-period folding to `[0, π/2)` gives a 2¹⁵ LUT with 2× headroom and works for arbitrary sequence length (table spans one quadrant, independent of positions).

## Acceptance criteria

- [ ] Per-scale trig period modulus derived by brute-force search (replaces `FOUR_PI_APPROX`); reproduces `3217` at s=8.
- [ ] cos/sin LUT quantized at `2^scale`.
- [ ] Quadrant folding so the LUT stays ≤ 2¹⁶ at scale 14 (≈ 2¹⁵).
- [ ] Remove the scale≠8 guard in `handle_cos`/`handle_sin`.
- [ ] cos/sin unit tests at scale 14 (mirroring `test_tanh_scale_12`); existing scale-8 tests still pass.
- [ ] Qwen e2e runs at scale 14 with cos/sin proven.

## Context

Follow-up to the rsqrt + tanh/erf/sigmoid scale-awareness work (gpt2 now verifies at scale 12). cos/sin were deferred because of the accuracy-vs-table-size tradeoff above.

AntoineF4C5:
This Pull Request implements an intermediate step in trigonometric operators proving (Cos/Sin) to downscale the teleported input, so that it can be fed into a smaller lookup table.

## Context
Since the lookup table for trig ops admits no prefix-suffix decomposition, its cost is highly correlated to the number of variates it uses. Additionally, we measured that precision of the cos/sin provers can be lowered without affecting model's precision (those operators are "too" precise).

Reducing the size of the lookup table is both a gain in prover/verifier cost, and can be done, to a certain extent, with no cost concerning precision.

## Proposal
Add an intermediate downscaling sumcheck between the remainder sumcheck and Trig lookup sumcheck, which reduces bit-width of the remainder to allow it to be sent to a smaller Trig lookup table.

This currently adds a 32-bits prefix-suffix shout lookup, and ceils the small-table number of vars to 16 (from 22 at scale 14).

## Performances

Node proving, still at scale 14, got significant results:
| |  Before  |   After  | 
|----|----|----|
| Prove  | 1.117s   | 633.115ms |
| Verify | 72.853ms | 8.190ms   |

## Running QWEN
We can now run QWEN with high output fidelity.

QEA (quant error analysis) for Qwen:
```
cargo run --release --package atlas-onnx-tracer --example qwen_quant_error_analysis
```

Prove + verify (jolt-atlas-core):
```
cargo run --release --package jolt-atlas-core --example qwen (-- --input "<Custom Input text>")
```

## Notes 
Results might differ at scale 12, namely because the natural number of vars small-tables take is already <= the bound `MAX_CUSTOM_TABLE_VARS` (16)

## Follow up work
- #279 would allow to get rid of the 32-bits shout lookup added by this PR while keeping the positive outcome.
- Applying the change to other teleported ops (activation nodes)
- Benchmarking precision and performance results for different `MAX_CUSTOM_TABLE_VARS` and different scales, to get best fit for each.

---

Closes #273 , closes #216.


Forpee (approved):
LGTM

Issue #283, opened 2026-08-19
Explore reducing the prover overhead of Saturation


AntoineF4C5:
## Context

Pull request #256 increased the VM soundness by constraining outputs of arithmetic operators to stay within i32. This had the effect to triple the proving time, tested for microgpt (this is at least as high overhead for gpt2).

## Benchmark

__Before (8fcd92c, parent of #256):__
- prove: avg 314.1ms (359.4 / 288.9 / 293.9ms)
- verify: avg 24.3ms (24.2 / 24.1 / 24.5ms)

__After (80a718e, #256 merged):__
- prove: avg 1061.6ms (1051.9 / 1062.2 / 1070.7ms)
- verify: avg 76.5ms (77.8 / 74.6 / 77.1ms)

Overhead:
- prove: +747.5ms, ~3.4x
- verify: +52.2ms, ~3.2x

## Idea

The saturation table has been implemented as always a 64-variates table proven via a 64-rounds shout sumcheck.

In most case, such a large table is not required:
- Add, Sub: output is perfectly represented in 33 bits, so any table size >=33 is enough.
- Mul, Square, ... : Raw output in 64bits, but we apply a rescaling before saturating: so a `64-Scale`-sized table is enough.
- Einsum, ... : Same idea, but we multiply and sum: `64 - Scale + log(reduced_dims)`-sized table (can be complex, we can skip optimization for this node now.

Investing on the gpt2 example showed that nodes Add, Mul and Einsum there share 90% of total saturating overhead, with roughly 30% each.

While the fix on Add could substentially reduce this overhead, Mul and Einsum might require an other trick to reduce the saturating overhead (like using smaller bit-width for the inputs).

PR #288, opened 2026-08-30
Perf/round 2

Forpee:
# Qwen prover: 222.8 s → 47.0 s (4.7×), plus a batch-opening soundness fix

Proves the Qwen example (9-token prompt) in **47.0 s** on an M3 Pro (11 cores), down from
222.8 s on `main` — **5.2 s per prompt token**, proof size 4.57 MiB (committed polynomials
18.1k → 3.4k). Verification passes throughout; core suite 116/116 + joltworks suite green.

⚠️ **Contains a soundness fix that should land regardless of the perf review** (happy to
split it out): see "Batch opening reduction fix" below.

## Part 1 — restructuring the node loop (222.8 → 81.0 s)

- **Deferred, batched lookups**: clamp/rescale lookups leave the node loop; claims are
  appended per node, the ~750 per-node sumchecks run afterwards as a few large batched
  sumchecks with parallel per-instance messages.
- **Packed lookup buckets**: nodes of the same clamp/rebase width share one cycle space
  (`CycleWeight::Packed`), so a bucket has one committed chunk set and one check set
  regardless of node count. SRS size unchanged.
- **Static clamp widths** (`{40,48,56,64}`) derived from sound per-op accumulation bounds.
- **Batched-GEMM einsum tracer kernel** (`Model::trace` 16.7 → 3.3 s) and fused-intermediate
  caching.
- **Dory tactical**: batched-affine tier-1 row sums, identity-row skipping in tier 2.

## Part 2 — protocol changes (81.0 → 47.0 s)

- **Interior/saturated clamp split** (`clamp_split.rs`): `out = SatClamp(acc)` is proven by a
  split sumcheck over `OUT = out + 2^31`, a virtual saturation indicator, and a sparse slack
  vector — 8 dense 32-bit chunks + `w/4` almost-empty slack chunks per bucket instead of
  `w/4` dense chunks and a degree-`w/4+1` RaVirtual check.
- **Exact modes**: a model *output* node whose public output has no saturated element, and
  any bucket the prover declares (and the verifier checks) unsaturated, prove `acc = out`
  directly from existing openings plus a degree-2 range check on `OUT`; slack chunks are
  committed empty and never opened. Data-dependent with a sound fallback: saturated buckets
  keep the split sumcheck.
- **Hamming weight folded into Booleanity**: `BooleanitySumcheckParams` gains an optional
  linear term `Σ_d (γ'_d + w_d·k)·eq(r_cycle,t)·ra_d(k,t)` sharing its G/H tables, so every
  one-hot chunk is opened exactly once (2 instances per group instead of 3).
- **Remainder buckets via value sumcheck**: the packed rescale-remainder range check drops
  its identity read-raf and RaVirtual.
- **Grouped openings**: polynomials appended together at one point reduce as a single
  instance on `Σ ρ^i·P_i` (reduction instances 10.2k → 3.6k).
- **Parallel Dory fold**: `dory-pcs`'s vector routines in the reduce-and-fold rounds are
  sequential loops; `ParG1Routines`/`ParG2Routines` parallelize them (joint open 12 → 5 s).
- Deferred softmax stages 3–4, eval-reduction single-opening short path, small-scalar
  einsum operand folds, misc.

## Batch opening reduction fix

`ProverOpeningAccumulator::sumchecks` (and the verifier's) was keyed by `CommittedPoly`, so a
polynomial opened at several points kept only its **last-appended** opening — for every
one-hot chunk, only the Booleanity opening was bound to its commitment; the RaVirtual and
Hamming openings never were (measured on a small model: 192 committed openings, 96 reduction
instances). Fixed by keying instances on `OpeningId` and deduplicating the reduced claims per
polynomial (all instances of a polynomial evaluate it at the same joint point). Proof format
unchanged. This gap exists on `main`; the fix costs prover time (~8 s here), which later
commits recover.

## Notes / caveats

- Benchmarked end-to-end on the Qwen example only; the test suite exercises every op path
  (including the exact-mode fallbacks via tampering tests), but gpt2/bge have not been
  re-benchmarked on this branch.
- `--features zk` does not build — pre-existing on `main` (23 errors before this branch).
- Per-phase numbers in commit messages before the soundness fix were measured with the gap;
  the 47.0 s headline includes the fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


atlas-onnx-tracer/src/ops/mod.rs
```diff
@@ -157,10 +157,57 @@ define_operators! {
 }
 
 /// Trait for all operators - defines how to execute the operation on input tensors.
```
AntoineF4C5, line 159: This doc should be moved back to just above "Op" trait.

jolt-atlas-core/src/onnx_proof/clamp_lookups/mod.rs
```diff
@@ -275,77 +276,31 @@ pub fn prove_clamp_lookup<F: JoltField, T: Transcript>(
...
-        (
-            ProofId(node.idx, ProofType::RaOneHotChecks),
-            ra_one_hot_proof,
-        ),
-    ]
+    prover.deferred.push(ProverLookupJob::Clamp {
+        node: node.clone(),
+        lookup_bits,
+    });
+    Vec::new()
```
AntoineF4C5, line 302: We can change this method's signature since it always returns an empty Vec.

Forpee:
to test: 
```
python scripts/download_qwen.py   # once, if the model isn't already there

cargo run --release --package jolt-atlas-core --example qwen
```

AntoineF4C5:
Closes #283

jolt-atlas-core/src/onnx_proof/clamp_lookups/mod.rs
```diff
@@ -320,18 +318,12 @@ pub fn verify_clamp_lookup<F: JoltField, T: Transcript>(
...
-        .instruction_d;
-    (0..d)
-        .map(|i| CommittedPoly::ClampRaD(node.idx, i))
-        .collect()
+/// A clamped node commits **no** per-node polynomials: its clamp one-hots live
+/// in a packed bucket (see [`global_clamp`](crate::onnx_proof::global_clamp)),
+/// whose chunks are listed at the model level by
+/// `global_clamp::bucket_committed_polys`.
+pub fn clamp_committed_polys(_node: &ComputationNode) -> Vec<CommittedPoly> {
+    Vec::new()
```
AntoineF4C5, line 344: We could remove this method as it returns an empty vec.

jolt-atlas-core/src/onnx_proof/deferred_lookups.rs
```diff
@@ -303,6 +391,39 @@ pub fn prove_all<F: JoltField, T: Transcript>(
...
     }
+
+    // Instances built during the node loop, batched here; each batch is
+    // followed by the one-hot checks that consume its `ra` claims.
+    let mut batches = std::mem::take(&mut prover.deferred_batches);
+    let onehots = std::mem::take(&mut prover.deferred_onehots);
+    for kind in [
+        DeferredBatch::ActivationExec,
+        DeferredBatch::ActivationClamp,
+        DeferredBatch::RangeCheck,
```
AntoineF4C5, line 486: It is better to iterate accross all variants of "DeferredBatch" for easier composition as we eventually populate this enum with new variants.

jolt-atlas-core/src/onnx_proof/prover.rs
```diff
@@ -49,6 +49,17 @@ pub struct Prover<F: JoltField, T: Transcript> {
     /// Lookups registered during the node loop, proven afterwards in one batch
     /// (see [`deferred_lookups`](crate::onnx_proof::deferred_lookups)).
     pub deferred: Vec<crate::onnx_proof::deferred_lookups::ProverLookupJob>,
+    /// Sumcheck instances built during the node loop and proven afterwards,
+    /// one batched sumcheck per [`DeferredBatch`](crate::onnx_proof::deferred_lookups::DeferredBatch).
+    pub deferred_batches: BTreeMap<
+        crate::onnx_proof::deferred_lookups::DeferredBatch,
+        Vec<Box<dyn joltworks::subprotocols::sumcheck_prover::SumcheckInstanceProver<F, T>>>,
```
AntoineF4C5, line 56: This comment serves as a general remainder to do a pass to import types instead of using fully qualified paths in case such as this where the type becomes really long.

joltworks/src/poly/opening_proof.rs
```diff
@@ -225,7 +230,11 @@ where
     /// Caches an opening claim from the opening reduction sumcheck.
     /// Called from `OpeningProofReductionSumcheckProver::cache_openings`.
     pub fn cache_opening_reduction_claim(&mut self, polynomial: CommittedPoly, claim: F) {
-        self.cached_opening_claims.insert(polynomial, claim);
+        // Every opening of `polynomial` reduces to its evaluation at the same
+        // joint point, so repeated claims must agree.
+        if let Some(prev) = self.cached_opening_claims.insert(polynomial, claim) {
+            debug_assert_eq!(prev, claim, "reduced claims of {polynomial:?} differ");
+        }
```
AntoineF4C5: A better fix to the issue identified in commit 2ce4b35 would be to perform an `EvalReduction` (curve-based opening reduction for several openings of a unique polynomial) to ensure there is a unique opening to each polynomial at time of performing opening_reduction. the EvalReduction requires no sumcheck is far less expensive than the method in opening_reduction, and the prover overhead of the fix should be better contained.

AntoineF4C5: I would also advocate for removing the optimization in commit 14ca88d if the proving time impact is negligible, since it introduces non-trivial complexity to a generic structure used accross the library, just for a localised edge-case (`LinearClaim`'s `terms` field namely).

jolt-atlas-core/src/onnx_proof/clamp_split.rs
```diff
@@ -137,6 +137,54 @@ fn ind_id(bucket_idx: usize) -> OpeningId {
...
+            .output_indices
+            .iter()
+            .position(|&i| i == node_idx)
+            .and_then(|p| io.outputs.get(p))
+            .is_some_and(|t| unsaturated(t.data()))
+}
+
+/// `bucket` restricted to the nodes that are proven through it (`live`);
+/// `None` if no node is.
+pub fn live_bucket(
```
AntoineF4C5, line 194: I am not a big fan of the way the "output node unsaturated shortcut" is implemented. It looks like it is adding a few complexities and "nested exception handling" at several places in the code base, for something that is exceptional and for which I would have hoped for a "toggle" or something similar, that disables or removes a claim if it is on the output node's saturation check in this case.

I do not have a precise way how I would have implemented that, but this reminds me of a discussion I had seen: https://stackoverflow.com/questions/4369822/early-returns-vs-nested-positive-if-statements
Not completely related, but I hope that explains my point a bit better.

jolt-atlas-core/src/onnx_proof/deferred_lookups.rs
```diff
@@ -294,33 +294,80 @@ pub fn prove_all<F: JoltField, T: Transcript>(
...
         );
 
-        // Challenges are drawn sequentially in bucket order; the (heavy)
-        // instance construction then runs in parallel across buckets.
-        let params: Vec<SplitParams<F>> = buckets
+        // Each bucket first declares (as a public claim) whether it has any
+        // saturated element; unsaturated buckets are proven exactly (value
+        // sumcheck on OUT + per-node `acc = out` on the verifier), the others
+        // by the split sumcheck. Challenges are drawn sequentially in bucket
+        // order; the (heavy) instance construction then runs in parallel.
```
AntoineF4C5, line 303: This commit adds a flag to each accumulated node, indicating whether accumulated values from this node saturate. If they do not, a shortcut is taken, the prover and verifier just need to assert that `acc = out`, and `acc` values are in `i32`.
The prover needs to send the flag values to the verifier in the clear, since the sumcheck verified differs depending on that value.

This means we can not hide the flag values through blindfold, as the checks performed are different depending on the sumcheck instance being run.

Despite being small, this is an information leakage as the verifier, even through blindfold, learns which node has saturating elements and which does not.

I think the information leakage is quite contained and should not provide much relevant knowledge to the verifier. Given the proving time improvements, we might still want to include this change.

joltworks/src/subprotocols/opening_reduction.rs
```diff
@@ -506,10 +589,61 @@ impl<F: JoltField> EqCycleState<F> {
 ///   \sum eq(k, r_address) * eq(j, r_cycle) * ra(k, j)
 /// so we use a simplified version of the prover algorithm for the
 /// Booleanity sumcheck described in Section 6.3 of the Twist/Shout paper.
```
AntoineF4C5, line 591: This doc should be moved back to just above `OneHotPolynomialProverOpening` struct.

joltworks/src/poly/opening_proof.rs
AntoineF4C5: The follow-up polynomial batching commit 4474a7f5840aa26ebbec02e6552bede01456ae2f should be reassessed after switching to `EvalReduction` for the fix, as the number of polynomials and their opening point may vary, making the polynomial batching optimization an added cost rather than a time profit.

AntoineF4C5 (approved):
Great Pull Request that gets substential improvements to the codebase.
Most of the implementation felt easy to read and understand, but the last few commits were quite harder to understand, with an implementation that could feel (over)complicated.
I think the great time savings allowed by AI in the implementation of new features should be used in making each feature as nice and simple as possible, reducing the implementing/reviewing spent-time asymmetry by both spending more time implementing, and allowing the review to be done faster.
On top of that, I fear that soon, a codebase where a lot of complexity and "edge-case handling" is added can become impossible to maintain even by AI, making for an other reason to spend time improving code simplicity.

I left a few comments about fixes of current PR and follow-ups to consider for further improvements.

jolt-atlas-core/src/onnx_proof/clamp_lookups/mod.rs
AntoineF4C5: Kept as useful as the comment gives context as to why we return an empty Vec in this case.

PR #290, opened 2026-09-16
Use sequential tensor iterators in RISC-V guests

Acentelles:
## Problem and change

Tensor helpers can initialize Rayon's global worker pool when Atlas verification runs in a Jolt RISC-V guest. The guest cannot start those threads, so verification fails with `ThreadPoolBuildError` and `EAGAIN` even when the host sets the parallel chunk size to disable splitting.

Use the existing WebAssembly iterator fallback for `riscv64` too, including the sequential dot-product branch. Move that fallback into a small module so its behavior can be tested on a host. Other host targets retain their existing parallel implementation.

## Validation

- Both fallback regressions pass. They cover borrowed and owned iteration with `Rc` values, mutable vector and slice iteration, and sorting.
- A native smoke executable linked against this Atlas revision imports the tracer and checks a dot product with positive and negative entries. It passes using the experiment's pinned Tract revision and arkworks compatibility export.
- The same fallback implementation was applied to the paper-pin Atlas export used by a Jolt verification guest. The rebuilt RISC-V guest completes the constant-preprocessing fixture and returns `true` after 86,143,886 cycles. The earlier implementation failed while initializing Rayon. This is a guest execution test, not an outer proof result.

The RISC-V integration check uses that exported verifier dependency with its existing ONNX import feature gate. This PR does not claim that the entire Atlas main workspace, including ONNX import and forward execution, now builds or runs in a guest.

The fallback regressions can also be run without building the full workspace:

```sh
rustc --edition 2024 --test atlas-onnx-tracer/src/utils/sequential_utils.rs -o /tmp/atlas-sequential-tests
/tmp/atlas-sequential-tests
```


Forpee (approved):
LGTM

PR #293, opened 2026-09-16
Preserve logical element order across padded reshapes

Acentelles:
## Topic and evidence summary

**Review group:** Correctness and verifier portability. **Type:** Bug fix.

Preserve real-element order when reshape dimensions are padded; lower affected cases through existing checked operators.

**Baseline and observed effect.** No timing or memory gain claimed. Qwen last-logit cosine on one prompt: 0.931922 → 0.999874, with the reference next token and top ten restored. This is numerical correctness evidence, not translation accuracy or a proving-speed comparison; the registered imported graph changes.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `a78b7a5c83ae`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #292, #294, #335.

## Problem and change

Padding each reshape dimension independently can change logical element order. For example, `[2, 3, 4] -> [6, 4]` becomes `[2, 4, 4] -> [8, 4]`; the fourth row of the first group is padding, so a flat reshape shifts the second group's data. The same issue misaligns Qwen2-0.5B's 14 query heads with its two groups of seven shared-key heads.

Detect reshapes that move real elements under padding and lower them to existing reshape, gather, broadcast and multiplication operators. A common trailing shape stays together, so attention gathers entire blocks instead of individual scalars. A mask clears newly introduced padding. Compatible reshapes retain their existing path. Update input/output identities and original-weight shadow mappings after inserting nodes. This changes the imported graph and its registered model identity; it adds no proof operator or unchecked verifier shortcut.

## Validation

- Eight reshape cases cover merging, splitting, non-power-of-two trailing dimensions, unequal padded element counts and the actual Qwen attention shapes. Compatible layouts remain unchanged.
- Three small ONNX fixtures pass imported forward and original-weight shadow checks.
- Complete ordinary Atlas proofs for those three imported fixtures verify and reject an altered output. Serialized proofs are 4,734, 4,734 and 5,206 bytes. These are correctness fixtures with test setup.
- The unchanged Qwen2-0.5B export, scale 14, sequence length 32, on one recorded prompt now selects the original ONNX Runtime next token (20728), with top-ten overlap 10/10. Last-logit cosine improves from 0.931922 to 0.999874 and mean absolute error is 0.0518575.
- The original-weight floating shadow after the correction matches the original ONNX reference with mean absolute error 0.00000289 and top-ten overlap 10/10, the same result as disabling padding.

The Qwen comparison retains only the final vocabulary projection after all transformer layers. A separate equality check confirms that this projection rewrite preserves every final logit. These are numerical checks on one prompt, not a translation-accuracy or full Qwen proof claim. Attention-mask headroom is tracked separately in #292. The Qwen proof measurement will use both changes.

Local builds use Rust 1.97.1, the pinned Tract revision `c484b3ee9a22e7d2bfca8394619771397b61c0d6`, and the same arkworks compatibility export used for the paper experiments. CI's native ZK module has existing API mismatches also seen on the other isolated branches; this change does not repair that module.


Acentelles:
Consolidated into https://github.com/ICME-Lab/jolt-atlas/pull/369.

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `a78b7a5c83ae0dca71e671658c8aaea87d5e75bb` remain available.

PR #295, opened 2026-09-16
Batch HyperKZG commitment combination into one MSM

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance + input-validation hardening.

Replace independent HyperKZG scalar multiplications with one exact MSM. Separately reject unequal base/scalar lengths instead of silently truncating zip.

**Baseline and observed effect.** Complete corrected Qwen guest: 61,320,942,771 → 42,719,182,959 cycles; 18,601,759,812 cycles less (30.34% reduction; 1.435× baseline/candidate). Baseline already uses inline Blake2; the same saved witness is used. Add/ReLU: 142,355,599 → 102,651,989 cycles. The outer fixture proof is accepted, but its wall time is not a matched proving comparison.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `8a11975ddc47`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #296, #297, #298, #299, #300, #301, #302, #309, #310, #311, #312, #313, #314.

## Summary

`HyperKZG::combine_commitments` currently computes each scalar multiplication separately and then sums the results. A measured current Qwen proof contains 18,154 commitments, making this expensive inside a verifier guest.

Use the existing `msm_field_elements` routine to compute the same linear combination in one MSM (a sum of points multiplied by scalar coefficients). This retains the points, coefficients, transcript and proof encoding. Mismatched lengths fail explicitly instead of silently truncating through `zip`.

## Validation

- Complete prepared-verifier regressions pass with the change, including malformed encodings, changed commitments, claims and execution IO.
- The same saved corrected Qwen proof accepts and all eight rejection controls pass.
- On the same Add/ReLU verifier fixture and Jolt guest configuration with Blake2 compression enabled, execution falls from 142,355,599 to 102,651,989 cycles (27.9%). Both executions accept the same saved witness. These are guest execution measurements, not outer proof timings.
- Added direct equality regressions against separate scalar multiplications for empty and mixed batches, repeated/opposite/identity points, and zero, unit, negative and random coefficients. A separate regression checks coefficient length mismatch.

The fixture now has a complete outer Jolt proof of 97,814 bytes, separately measured verifier preprocessing of 67,347 bytes and 160 bytes of public IO. Fresh verification accepts, rejects a changed statement/result, and confirms private advice is absent from serialized IO. Proving took 394.48 seconds after 441.02 seconds of fresh setup on the EPYC 7B13 with 32 Rayon threads. This is one measurement with other traces running, not a controlled timing comparison.

The complete corrected Qwen guest now accepts in 42,719,182,959 cycles, compared with 61,320,942,771 with inline Blake2 before the MSM change, a 30.3% reduction on the same saved witness. Both counts remain above the current guest proving bound. No Qwen outer proof is claimed.


Acentelles:

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `8a11975ddc47ebf388a3edad7c2e47e3902558a7` remain available.

PR #296, opened 2026-09-16
Evaluate univariate polynomials with Horner rule

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Use Horner evaluation for ordinary and compressed univariate polynomials, including reconstruction of the omitted coefficient.

**Baseline and observed effect.** Complete corrected Qwen guest: 42,719,182,959 → 42,618,879,428 cycles; 100,303,531 cycles less (0.23% reduction; 1.002× baseline/candidate). Baseline includes #295 and Blake2 compression. This is guest execution; outer proving time is unmeasured for this isolated change.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `8f23c44a62aa`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #297, #298, #299, #300, #301, #302, #309, #310, #311, #312, #313, #314.

## Summary

Univariate evaluation currently constructs powers of the evaluation point and multiplies each coefficient by those powers. Evaluate with Horner's rule instead, including compressed sumcheck polynomials after reconstructing the omitted linear coefficient. For degree d, this uses d multiplications by the point and avoids the additional products by its powers. Challenge types keep their existing optimized multiplication path.

This changes arithmetic order only. Coefficients, transcript messages and serialized proofs are unchanged.

## Validation

- Added comparisons with independent power-basis evaluation for coefficient counts 1, 2, 3, 4, 8, 17 and 64, including leading zeros, zero/unit/maximal/random challenges, full field points and arbitrary reconstruction hints. The regression passes in a standalone dependency harness.
- A complete native ReLU proof verifies and a changed output is rejected. Its 7,804 serialized bytes are identical to the saved baseline fixture proof.
- The complete saved corrected Qwen proof accepts with this change and all eight rejection controls pass. The prepared witness bytes are unchanged.
- The same RISC-V fixture accepts with 102,502,937 cycles, compared with 102,651,989 before this change (both already include batched commitment combination and Blake2 compression). The complete corrected Qwen guest accepts in 42,618,879,428 cycles, down from 42,719,182,959 before this change (0.235%). This remains above the configured outer proving bound; no Qwen outer proof is claimed.

Issue #289, opened 2026-09-16
Restore ZK operator support and add a Qwen prove/verify regression


Acentelles:
## Problem

The current Qwen example uses the ordinary Dory prover. A successful run of that example does not exercise the BlindFold ZK pipeline. Downstream Qwen experiments need a working ZK prove/verify path with an explicit statement of what remains public.

This is a focused follow-up to #28, based on source inspection of `main` at `bb9e1e6d215eaccef233fc38faea1bb3519e7365`. The existing [Qwen example](https://github.com/ICME-Lab/jolt-atlas/blob/bb9e1e6d215eaccef233fc38faea1bb3519e7365/jolt-atlas-core/examples/qwen.rs#L127-L142) calls `ONNXProof::prove`; the ZK implementation currently fixes HyperKZG as its PCS. The ZK example will need matching preprocessing.

## Current gaps

All references below are pinned to the inspected revision.

| Operator or path | Missing implementation |
|---|---|
| `SoftmaxLastAxis` | The clamp-based flow has explicit ZK prover and verifier stubs. [Prover](https://github.com/ICME-Lab/jolt-atlas/blob/bb9e1e6d215eaccef233fc38faea1bb3519e7365/jolt-atlas-core/src/onnx_proof/zk.rs#L1022-L1042), [verifier](https://github.com/ICME-Lab/jolt-atlas/blob/bb9e1e6d215eaccef233fc38faea1bb3519e7365/jolt-atlas-core/src/onnx_proof/zk.rs#L285-L299). |
| `MeanOfSquares` | Both dispatch paths explicitly reject the missing ZK implementation. [Prover](https://github.com/ICME-Lab/jolt-atlas/blob/bb9e1e6d215eaccef233fc38faea1bb3519e7365/jolt-atlas-core/src/onnx_proof/zk.rs#L1858-L1863), [verifier](https://github.com/ICME-Lab/jolt-atlas/blob/bb9e1e6d215eaccef233fc38faea1bb3519e7365/jolt-atlas-core/src/onnx_proof/zk.rs#L2740-L2746). |
[...]

PR #297, opened 2026-09-16
Evaluate unpadded reshape selectors without materializing tensors

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Replace materialized unpadded reshape selectors with an equality-polynomial formula; padded cases retain the table path.

**Baseline and observed effect.** Complete corrected Qwen guest: 42,618,879,428 → 36,241,922,278 cycles; 6,376,957,150 cycles less (14.96% reduction; 1.176× baseline/candidate). Baseline includes #295–#296; same witness and Blake2. Domain checks protect the fast path; no prior acceptance bug is asserted.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `6d2e47dd0d58`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #298, #299, #300, #301, #302, #309, #310, #311, #312, #313, #314.

## Summary

Reshape verification constructs a selector table with one entry per padded tensor element, then evaluates that table at the input opening point. When every input and output dimension is a power of two, flat indices are unchanged. The selector evaluation is exactly the equality polynomial at the two opening points.

Evaluate that formula directly in both ordinary verification and the ZK challenge-value calculation. Keep the existing table path for shapes requiring padding. Check the tensor domain and opening-point lengths before taking the direct path. The prover, coefficients, transcript and proof encoding are unchanged.

## Validation

`cargo test --release -p jolt-atlas-core --lib onnx_proof::ops::reshape::tests` passes all four tests with the recorded local arkworks compatibility patches.

- Compare the direct result against the original materialized table on seven shape pairs, random and Boolean points, singleton dimensions, an empty opening point for a one-element tensor, and shapes with internal padding.
- Reject an opening point of the wrong dimension even when both supplied point lengths agree.
- Existing complete proof tests pass for four unpadded shape pairs and the padded `[10, 10]` to `[20, 5]` reshape.

The recorded corrected Qwen proof has 195 reshapes, all with dimensions that are powers of two. The complete guest accepts in 36,241,922,278 cycles, compared with 42,618,879,428 before this change (14.96% fewer). Both runs use the same prepared witness and already include PRs #295 and #296 and Jolt Blake2 compression. All eight native rejection controls pass. This remains above the configured outer proving bound, and no Qwen outer proof is claimed.

## CI scope

Native/default tests, GPT-2, BGE, formatting and WASM checks pass. The native ZK job fails at unchanged division and lookup API imports, also present on the other branches from this base. The direct selector formula is shared with ZK challenge calculation, but a complete native ZK regression is not established by this PR. Broader Qwen ZK coverage is tracked in #289.


Acentelles:

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `6d2e47dd0d582dc1b8cc39310049c782fda02a17` remain available.

PR #299, opened 2026-09-16
Evaluate aligned concat selectors without tensor tables

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Evaluate aligned concat selectors directly, retaining padding and unequal-input-size factors and the fallback path.

**Baseline and observed effect.** Complete corrected Qwen guest: 34,044,908,725 → 32,090,637,857 cycles; 1,954,270,868 cycles less (5.74% reduction; 1.061× baseline/candidate). Baseline includes #295–#298. No outer proving-time claim.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `7fefdef956ef`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #300, #301, #302, #309, #310, #311, #312, #313, #314.

## Summary

Concat verification constructs and evaluates a selector table for every input tensor. When an input occupies an aligned block in an output without padding, its selector is the product of fixed output-bit factors and an equality polynomial. Evaluate that product directly, including the zero factors for the extra variables used when inputs have different sizes.

Apply the formula in ordinary verification and ZK challenge-value calculation. Retain the original table path for unaligned blocks or shapes with padding. Shape and point-dimension checks remain explicit. The prover, transcript and proof format are unchanged.

## Validation

`cargo test --release -p jolt-atlas-core --lib onnx_proof::ops::concat::tests` passes all six tests with the recorded local arkworks compatibility patches.

- Compare every input selector against the original materialized table for eight shape configurations, random and Boolean points, different axes, unequal input sizes, unaligned blocks, internal padding, a single input and scalar inputs.
- Reject opening points of the wrong dimension.
- Complete proof tests cover ordinary and negative axes, mixed input sizes, both direct and fallback paths, and the identity case.

The recorded corrected Qwen proof has 49 concatenations, all meeting the direct-path conditions. The complete guest accepts in 32,090,637,857 cycles, compared with 34,044,908,725 before this change (5.74% fewer). Both runs use the same prepared witness and already include PRs #295–#298 and Jolt Blake2 compression. All eight native rejection controls pass. This remains above the configured outer proving bound, and no Qwen outer proof is claimed.

## CI scope

[...]

PR #298, opened 2026-09-16
Evaluate aligned slice selectors without tensor tables

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Evaluate aligned slice selectors through fixed-bit factors and equality weights, retaining fallback for unaligned or padded shapes.

**Baseline and observed effect.** Complete corrected Qwen guest: 36,241,922,278 → 34,044,908,725 cycles; 2,197,013,553 cycles less (6.06% reduction; 1.065× baseline/candidate). Baseline includes #295–#297. No outer proving-time claim.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `40f99419c2e7`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #299, #300, #301, #302, #309, #310, #311, #312, #313, #314.

## Summary

Slice verification materializes a selector with one entry per input element. For shapes whose dimensions are powers of two, a slice aligned to its output length fixes high bits of the sliced axis and preserves the other bits. Evaluate the corresponding fixed-bit factors and equality polynomial directly.

Use the formula in ordinary verification and ZK challenge-value calculation. Unaligned blocks and padded shapes retain the original table path. Shape validation and opening-point length checks remain explicit. The prover, transcript and proof format are unchanged.

## Validation

`cargo test --release -p jolt-atlas-core --lib onnx_proof::ops::slice::tests` passes all three tests with the recorded local arkworks compatibility patches.

- Compare with the original materialized table on eight cases, using random and Boolean points. Cases cover different axes, nonzero offsets, singleton output axes, the entire input, unaligned blocks and padding on the sliced or another axis.
- Reject points of the wrong dimension.
- Ten complete proof cases cover both direct evaluation and fallback paths.

The recorded corrected Qwen proof has 97 slices, all meeting the direct-path conditions. The complete guest accepts in 34,044,908,725 cycles, compared with 36,241,922,278 before this change (6.06% fewer). Both runs use the same prepared witness and already include PRs #295–#297 and Jolt Blake2 compression. All eight native rejection controls pass. This remains above the configured outer proving bound, and no Qwen outer proof is claimed.

## CI scope

[...]

PR #300, opened 2026-09-16
Reuse equality weights when verifying softmax max indicators

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance + input-validation hardening.

Share softmax column equality weights or use direct bit factors according to the bounded table-size choice; reject invalid rows and indices.

**Baseline and observed effect.** Complete corrected Qwen guest, #300 and #301 together: 32,090,637,857 → 31,744,903,156 cycles; 345,734,701 cycles less (1.08% reduction; 1.011× baseline/candidate). Baseline includes #295–#299. The two patches are not isolated from one another; neither individually owns the full 1.08% reduction.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `26c18315044d`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #299, #301, #302, #309, #310, #311, #312, #313, #314.

## Summary

The softmax max-indicator verifier builds a field-valued bit vector and evaluates the column equality polynomial independently for every row. Reuse a column equality table when its size is at most the cost of visiting every row's index bits. For wider tables, select the zero/one factors directly without allocating per-row bit vectors.

Both paths evaluate the existing sparse indicator relation exactly. The table choice bounds the added memory by the existing row/index work. Apply the shared helper to ordinary verification and ZK challenge values, with explicit row-count and index bounds. The prover, transcript and proof format are unchanged.

## Validation

The `onnx_proof::ops::softmax_last_axis` native test group passes four tests; the existing scale-twelve-only test remains ignored at the default model scale.

- Compare against both dense multilinear evaluation and the original sparse bit-vector formula on six row/column dimensions, with repeated and random indices, random field points and Boolean points.
- Exercise both table reuse and direct bit-factor paths, including a singleton domain.
- Reject a wrong row count and out-of-range indices.
- The existing complete softmax proof test at the configured model scale passes.

These tests used the recorded local arkworks compatibility patches. Full corrected Qwen native checks and guest profiling are separate measurements; no model-scale speedup is claimed yet.

## Combined Qwen measurement

With PRs #295–#299 already applied, adding #300 and #301 together reduces the complete corrected Qwen guest from 32,090,637,857 to 31,744,903,156 cycles (1.08%). The saved witness is unchanged and all eight native rejection controls pass. This measurement does not separate the two changes. It remains above the configured outer proving bound, with no outer Qwen proof.


Acentelles:

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `26c18315044dd096da5814caaa3c57395bbb3222` remain available.

PR #310, opened 2026-09-16
Evaluate reduction interpolation with a shared Lagrange basis

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Reuse a Lagrange basis instead of solving for coefficients per coordinate; the later addition reads stored endpoint values at zero and one.

**Baseline and observed effect.** Prepared grouped Qwen verifier: 7,112,443,367 → 3,894,564,434 cycles; 3,217,878,933 cycles less (45.24% reduction; 1.826× baseline/candidate). Baseline already has #309 and #311. The 45.24% result applies to shared interpolation; do not separately assign it to the later endpoint shortcut.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `e7f090564995`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #299, #300, #301, #302, #309, #311, #312, #313, #314.

Evaluation reduction currently interpolates a coefficient polynomial for every coordinate before evaluating it at one challenge. At nodes with many consumers, the repeated Gaussian elimination is expensive.

Compute the Lagrange basis on the shared integer grid once, using prefix/suffix products and one factorial inversion, then evaluate all coordinates with those weights. The short paths remain direct. This preserves the interpolation polynomial, proof checks and transcript. It also works when the challenge equals a grid point.

Validation passes all 11 evaluation-reduction tests, including an independent comparison with coefficient interpolation on grids up to 48 points, every grid point, random challenges, transcript mismatch and altered-proof rejection. The complete saved Qwen proof accepts with the same witness hash and all 12 missing-deferred-proof controls plus the existing eight tamper controls.

In the zkARc prepared verifier with shared activation tables and fixed lookup tables, this change reduces actual Jolt execution from **7,112,443,367 to 3,894,564,434 cycles (45.2%)** on the identical saved proof. Both guests use Rust 1.95, release overflow checks and inline Blake2. This measures execution, not an outer proof or native proving time. Reproduction sources and receipts are published under `ICME-Lab/zkARc`, `round2-recursion/` and `bench-logs/2026-09-16-round2-recursion/`.

[...]

PR #309, opened 2026-09-16
Share activation tables and evaluations during verification

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Share activation tables within verification and cache evaluations only for the same table type and complete evaluation point.

**Baseline and observed effect.** Prepared grouped Qwen verifier: 23,032,033,118 → 8,109,826,681 cycles; 14,922,206,437 cycles less (64.79% reduction; 2.84× baseline/candidate). Same saved proof; Rust 1.95, overflow checks and inline Blake2. This isolates the cache in the integration, not outer proving.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `8718b9923500`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #299, #300, #301, #302, #310, #311, #312, #313, #314.

Clamped activation nodes currently materialize a separate fixed table and evaluate it for every verifier instance. Qwen has 24 sigmoid instances, each with 262,144 table entries. Deferred instances share the same evaluation point.

This change shares each table within one verification and retains its last evaluation. The cache distinguishes table marker types and compares the complete point before reuse. Different points are recomputed. A deterministic map also works in a Jolt guest without operating-system randomness. Proof encoding, challenges and acceptance equations are unchanged.

Validation includes independent dense equality sums for sigmoid, tanh and erf, changed evaluation points, and a complete proof with repeated and mixed activations. All pass. The complete saved Qwen proof also accepts with the same private witness hash, while removal of each of its 12 deferred proofs and the existing eight tamper controls is rejected.

In the zkARc prepared verifier built from the combined round-two integration, changing only this cache reduces actual Jolt execution from **23,032,033,118 to 8,109,826,681 cycles (64.8%)** on the same saved proof. Both guests use Rust 1.95, release overflow checks and inline Blake2. This is an execution measurement, not an outer proof or a native proving-time speedup. Reproduction sources and receipts are published in `ICME-Lab/zkARc` under `round2-recursion/` and `bench-logs/2026-09-16-round2-recursion/`.

Published reproduction and measurements: [zkARc checkpoint c1d5ed4](https://github.com/ICME-Lab/zkARc/tree/c1d5ed407c9ee8ee0ae86a95b40069b15a586faf/bench-logs/2026-09-16-round2-recursion).

PR #302, opened 2026-09-16
Evaluate equality factors with one field product

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Use 1-x-y+2xy instead of xy+(1-x)(1-y), removing one multiplication per equality factor.

**Baseline and observed effect.** The arithmetic work reduction is exact. No isolated full-Qwen, native prover, outer-prover or memory gain is established by the reported arithmetic regressions.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `bb4c7e87dd67`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #299, #300, #301, #309, #310, #311, #312, #313, #314.

## Summary

Each coordinate of the equality polynomial currently computes `x*y + (1-x)*(1-y)`. Evaluate the identical expression `1-x-y+2*x*y`, saving one field multiplication per coordinate. Apply this in the generic challenge-aware function and both bit-order paths.

The existing length checks, optimized challenge products, transcript messages and proof encoding are preserved.

## Validation

`cargo test --release -p joltworks --lib poly::eq_poly::tests::test_mle` passes both regressions with the recorded local arkworks compatibility patches.

- Compare with the original two-product formula at random field points, mixed field/challenge points, challenge pairs, and zero, one and negative one.
- Cover dimensions 0, 1, 2, 7, 16, 23 and 64, and equal and opposite bit orders.
- Retain rejection of unequal point lengths.

Both complete native reshape proof tests also pass, including a shape with padding. No Qwen speedup or outer proof size is inferred from the arithmetic regression.

PR #301, opened 2026-09-16
Interpolate linear polynomials directly

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance.

Replace the two-point Vandermonde solve with coefficients e0 and e1-e0, retaining the original trimmed representation.

**Baseline and observed effect.** Complete corrected Qwen guest, #300 and #301 together: 32,090,637,857 → 31,744,903,156 cycles; 345,734,701 cycles less (1.08% reduction; 1.011× baseline/candidate). Baseline includes #295–#299. The two patches are not isolated from one another; neither individually owns the full 1.08% reduction.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `e52b16e040db`, code base `29170bd3a813`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #299, #300, #302, #309, #310, #311, #312, #313, #314.

## Summary

`UniPoly::from_evals` has direct formulas for three and four evaluations, but two evaluations fall through to a Vandermonde matrix solve. For evaluations at zero and one, the coefficients are simply `e0` and `e1 - e0`.

Use that formula and retain `from_coeff` so constant and zero polynomials have exactly the same trimmed representation and serialized bytes. This also avoids repeatedly solving tiny systems when evaluation reduction interpolates each coordinate of a path through two points. No transcript or proof format changes.

## Validation

All eleven tests in `poly::unipoly::tests` pass with the recorded local arkworks compatibility patches. The new regression compares coefficients and serialized bytes against the original Vandermonde routine for zero, constant, unit, negative-slope and 128 random pairs, and checks both supplied evaluations. Full model measurements are separate; no speedup is inferred from this unit test.

## Combined Qwen measurement

With PRs #295–#299 already applied, adding #300 and #301 together reduces the complete corrected Qwen guest from 32,090,637,857 to 31,744,903,156 cycles (1.08%). The saved witness is unchanged and all eight native rejection controls pass. This measurement does not separate the two changes. It remains above the configured outer proving bound, with no outer Qwen proof.

PR #311, opened 2026-09-16
Embed validated fixed activation tables for verifier guests

Acentelles:
## Topic and evidence summary

**Review group:** Ordinary verifier arithmetic and caches. **Type:** Performance / optional feature.

Embed verified scale-14 activation values in the guest instead of generating them with floating-point operations.

**Baseline and observed effect.** Prepared grouped Qwen verifier: 8,109,826,681 → 7,112,443,367 cycles; 997,383,314 cycles less (12.30% reduction; 1.14× baseline/candidate). Baseline includes #309. ELF size increases from 7,407,016 to 8,981,352 bytes (+1,574,336 bytes); other scales and the default build retain generators.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `4d7b8324be38`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #295, #296, #297, #298, #299, #300, #301, #302, #309, #310, #312, #313, #314.

Add an optional `fixed-tables` feature that embeds the scale-14 sine, cosine and sigmoid lookup values instead of constructing them with floating point operations at runtime. These values are constants in the verifier program. Other compile-time scales retain the reference generators. Table evaluation, challenges and proof checks are unchanged.

The tables occupy 1.5 MiB before compression. A standalone Rust generator, hashes and a test comparing every entry with the original Atlas tensor functions accompany them. The default build keeps runtime generation. `check-fixed-tables` enables complete runtime comparisons for validation across targets.

Validation passes the full entry comparison on macOS, exact regeneration on Linux, and every entry comparison inside the real RISC-V Qwen verifier. The saved Qwen proof accepts in both the audit and optimized guests.

In the zkARc prepared verifier with shared activation tables, the feature reduces actual Jolt execution from **8,109,826,681 to 7,112,443,367 cycles (12.3%)** on the same proof. The ELF grows from 7,407,016 to 8,981,352 bytes. Both guests use Rust 1.95, release overflow checks and inline Blake2. The audit guest is separate and is not used for this performance figure. These are execution measurements, not an outer proof or native proving-time speedup.

Published reproduction and measurements: [zkARc checkpoint c1d5ed4](https://github.com/ICME-Lab/zkARc/tree/c1d5ed407c9ee8ee0ae86a95b40069b15a586faf/bench-logs/2026-09-16-round2-recursion).


Acentelles:

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `4d7b8324be388e22c5b40faa20ebeb144c00232a` remain available.

PR #318, opened 2026-09-17
Prove exact integer multiplication with rescaling and clamping

Acentelles:
## Topic and evidence summary

**Review group:** Native operator coverage and integer semantics. **Type:** New capability + scalar correctness fix.

Prove exact integer multiplication, floor rescaling and clamping; handle the empty equality polynomial for scalar proofs.

**Baseline and observed effect.** No matched performance improvement reported. Integer-domain and false-remainder/clamp checks establish the new relation. Scalar support should remain separately visible from the new multiplication capability.

**Dependencies.** This PR targets #317 (`feat/native-private-lookups`); reviewed head `61ac2436c61c`, code base `18241b3fdccb`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #320, #321, #322, #323, #324, #325, #326, #327, #328, #330, #331, #332, #333, #334, #336, #339.

Prove exact integer multiplication followed by floor division by a registered power of two and clamping to the signed integer range. Hidden range checks bind both inputs, output, remainder and clamp gaps. The native proof requires the Dory opening and BlindFold argument.

Scalar proofs use the empty equality polynomial correctly. Numerical tests compare actual Atlas kernels, including negative rounding and clamp endpoints. Consistent false remainders and clamp witnesses reject. The example saves a public bundle for verification in a separate process.

Depends on #317. The series starts from `perf/round-2`.

## Validation

- Rust 1.97.1, authenticated source export of `61ac2436c61c76af25d06f7da4d35ddb1bb1a536`. Workspace crates rebuilt for this branch.
- `cargo nextest` library suites pass 202 tests with `zk` and 159 without it.
- Library Clippy passes with warnings denied.
- Native receipt example Clippy passes with warnings denied.
- The core library compiles with `zk`.

These checks cover the native component API. Complete private ONNX integration, authenticated model IO and representation conversions remain open. Related to #289.

## Repository CI scope

The focused library checks above pass. Repository-wide ZK CI remains blocked by the existing ONNX integration issues tracked in #289, including activation tests that still refer to `ACTIVATION_TABLE_BOUND` and unused softmax ZK helpers. The activation-test references are also present on the pinned `perf/round-2` base. This PR does not establish a working private ONNX dispatcher or resolve that issue.

PR #340, opened 2026-09-18
Fix zero-variable split equality initialization

sambarnes:
`GruenSplitEqPolynomial::new(&[], order)` currently returns `Default`: a zero scalar, empty caches, and the default binding order. The zero-variable equality polynomial is the constant one.

Initialize that branch with scalar one, neutral cache tables, and the requested order. Nonempty construction and `new_with_scaling` are unchanged. No API or proof-format change.

The regression test covers both binding orders, the scalar, dimensions, cache accessors, and the merged polynomial. It fails on the original code.

Validation on `main` (`bb9e1e6`), Rust 1.88:

- `cargo test --locked -p joltworks --lib`: 157 passed.
- `cargo clippy --locked -p joltworks --lib --tests -- -D warnings`: passed.
- `cargo fmt --all --check`: passed.


sambarnes:
Closing as a duplicate of the empty-equality fix already included in #318, which also handles scaled empty inputs. No review needed here.

sambarnes:
apologies for the notif churn here!

i'm at openrouter and had a ~6mo zk phase while back at opensea a few years ago -- will post more on an issue for discussion 🙏

PR #339, opened 2026-09-18
Prove shifted greedy selection and stopping over hidden tensors

Acentelles:
## Topic and evidence summary

**Review group:** Native operator coverage and integer semantics. **Type:** New capability + soundness enforcement.

Prove shifted greedy selection, stopping and zero padding over original hidden tensors; project required score rows before selection.

**Baseline and observed effect.** No isolated baseline-to-candidate timing or memory improvement is established in this PR’s recorded comparison. The new exact relation and its rejection tests are the deliverable; structural work avoidance alone is not a measured model speedup.

**Dependencies.** This PR targets #338 (`feat/native-tensor-boundaries`); reviewed head `c634d95847ee`, code base `33e70b43e31e`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #318, #320, #321, #322, #323, #324, #325, #326, #327, #328, #330, #331, #332, #333, #334, #336.

Prove shifted greedy selection and a registered stopping rule over the original hidden token and score tensors. Each response token equals the first maximum of the preceding score row. End tokens are excluded before the final response token, and the last token must be the end token unless the registered generation limit was reached. All remaining token positions are proved zero.

Project only response prediction rows before the maximum and prefix-product relation. A registered score offset permits a model to project just those positions. Original tensor openings, logical vocabulary bounds and the first-tie rule are preserved.

Tests cover accepted serialized proofs, ties, wrong response tokens, early end tokens, unfinished responses, nonzero padding, invalid lengths, projected score offsets and substituted producer commitments. Two complete component fixtures pass fresh verification from saved files.

This is the selection and stopping component. Causal model execution and canonical text conversion remain separate required relations; independently supplied scores do not establish model provenance.

Depends on #338. The series starts from `perf/round-2`.

## Validation

- Rust 1.97.1, authenticated source export of `c634d95847ee02b9049891c70fe6e20bae2bcc22`. Workspace crates rebuilt for this branch.
- `cargo nextest` library suites pass 296 tests with `zk` and 164 without it.
- Library Clippy passes with warnings denied.
- Native receipt example Clippy passes with warnings denied.
- The core library compiles with `zk`.

These checks cover the native component API. Complete private ONNX integration, authenticated model IO and representation conversions remain open. Related to #289.

## Repository CI scope

[...]

PR #315, opened 2026-09-17
Compact ordinary Atlas proofs with bounded decoding

Acentelles:
## Topic and evidence summary

**Review group:** Proof transport and shared commitments. **Type:** Performance + decoder hardening.

Compress ordinary HyperKZG advice, identifiers and exact repeated coefficients; bound decoding and restore original canonical transcript bytes.

**Baseline and observed effect.** Complete ordinary Qwen inner proof: 5,188,295 → 2,941,543 bytes; 2,246,752 bytes less (43.30% reduction; 1.764× baseline/candidate). Prepared input: 6,535,393 → 4,288,657 bytes. Matched canonical/v2 guest cycles: 1,940,734,622 → 1,906,064,251 (1.79% fewer). Compared with v1, v2 saves 7.51% proof bytes but costs 0.062% more cycles; all pad to 2^31 rows. No measured outer proving or peak-memory gain. Malformed/canonicality checks harden the new format; this does not imply the old canonical format was unsound.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `b80c74bf533d`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #343.

Ordinary BN254/HyperKZG proofs repeat softmax identifiers, small advice values, and reduction coefficients. This adds an opt-in compact format with bounded direct decoding and exact recovery of the canonical proof bytes, including streaming those bytes into an existing transcript hash. The recorded complete Qwen proof shrinks from 5,188,295 to 2,941,543 bytes (43.30%).

The format stores contiguous softmax u32 advice in runs, shares ordinary identifier prefixes, and references exact one-coefficient reductions already present in opening claims. Decoding reconstructs every original claim and coefficient before normal verification. It rejects invalid scalar/point encodings, oversized collections, duplicate or unordered keys, nonmaximal prefixes/runs, alternative literal/reference forms, overflow, truncation and trailing bytes. Existing canonical serialization is unchanged. Native ZK and Dory formats are outside this API's scope.

[...]

PR #343, opened 2026-09-18
Share native range commitments and add checked compact GT transport

Acentelles:
## Topic and evidence summary

**Review group:** Proof transport and shared commitments. **Type:** Performance + new transport capability + decoder hardening.

Part A shares only graph-identical range commitments with fresh class randomness. Part B uses checked 192-byte GT encoding rather than 384 bytes; graph-derived ordering enables external identifier omission.

**Baseline and observed effect.** Part A: 57,838 → 49,062 distinct commitments; witness/commitments 764.53 → 661.79 s (13.44% lower); proving 1,129.25 → 1,136.95 s, combined 5.0% lower. Same EPYC 9654, 48 workers, one sample each. Part B plus the external message codec: native messages 23,377,721 → 9,663,363 bytes, complete receipt 23,904,522 → 10,190,139 bytes. Warm-file decode/key-decode/restored-write mean 28.64 → 7.70 s, two samples each in ABBA order, excluding verification. These are different scopes; do not attribute all receipt savings to one subcomponent. Original native proof body stays 243,411 bytes.

**Dependencies.** This PR targets #339 (`feat/native-greedy-sequences`); reviewed head `28a311643703`, code base `c634d95847ee`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #315.

Repeated range checks on the same registered tensor currently create fresh commitments to identical digit-indicator polynomials. This change commits once per graph-defined class and reuses its commitment and opening hint at the other registered uses. A separate opt-in GT transport wrapper stores each remaining commitment in 192 rather than 384 bytes. The graph also exposes its canonical required commitment order so transports can omit redundant identifiers.

The sharing key includes the original committed tensor, bit width, digit width, offset and digit position. Equality of private witness values never determines sharing. Separate tensor identities keep separate randomness, and each proof uses fresh class randomness. Every original range relation, sumcheck, transcript entry and opening remains. The native verifier and its statement/proof formats are unchanged.

`CompactDoryCommitment` uses the quadratic torus coordinate `t = b/(a+1)` for `x = a + b*w`. Decoding restores the exact Fq12 element and invokes the existing prime-order subgroup check. Norm-one membership alone is insufficient. The identity encodes as zero; the omitted order-two point is outside GT. The default commitment serialization and transcript representation remain unchanged. This follows the quadratic parametrization described in [Rubin and Silverberg, section 5.2](https://www.iacr.org/archive/crypto2003/27290348/27290348.pdf).

Depends on #339. The two commits separate range sharing from the compact transport API. The benchmark's model dispatcher and versioned message codec live outside this PR; all changed library files are identical in the measured source export and this branch.

Validation on the exact upstream branch: 300 ZK and 167 ordinary library tests, scoped Clippy with warnings denied, and core library compilation with `zk`. Tests cover fresh randomness across proof invocations, independent blinds for distinct tensors with equal values, changed alias rejection, exact GT round trips including identity/inverses, noncanonical fields and norm-one points outside the target subgroup.

The full bounded Qwen2-0.5B artifact passes the unchanged original native verifier with both original tensor boundaries and fresh canonical prompt/response conversion proofs. Structural commitment counts fall from 57,838 to 49,062 (15.2%). The complete compact receipt has also passed independent verification. Native messages fall from 23,377,721 to 9,663,363 bytes (58.7%); the complete receipt including every consumed boundary/conversion file falls from 22.8 to 9.7 MiB (57.4%). The native proof body remains 243,411 bytes. Linux and macOS reconstruct byte-identical original statements and proof bodies.

The measured codec export passes 309 ZK and 176 ordinary tests, scoped Clippy, the core `zk` check, and its transport test/Clippy. Full-message rejection checks cover non-subgroup torus coordinates, noncanonical fields, trailing/truncated messages, missing required proofs and swapped conversion statements. The codec's ordered vector retains each full subgroup check and permits Arkworks parallel validation. A same-host warm-file ABBA decode comparison on the saved full proof measures 28.64 versus 7.70 seconds (two samples each); this includes key decoding and restored-file writes, excluding native proof verification.

On the same EPYC 9654 with 48 workers, identical input and the same registered key, witness construction plus commitments falls from 764.53 to 661.79 seconds (13.4%). Native proving takes 1,129.25 versus 1,136.95 seconds; the two phases combined improve by 5.0%. Peak invocation RSS is 130.25 versus 126.36 GiB. These are single samples. Both native proofs pass fresh verification; the control's conversions are not reproved and it is not a second complete generation receipt.

Witness polynomials and opening hints still exist for every original ID. This change does not establish the separate application composition/simulation assumptions or a general reduction in peak memory. Scoped library checks do not claim that the pre-existing experimental ONNX ZK dispatcher passes full repository CI.

[Complete measurements, exact source pins and authenticated public proof/verifier archives](https://github.com/ICME-Lab/zkARc/blob/0fff02884732c9210deebe6b803c36cbc917efae/native-atlas/fresh-commitments/results.md) include reproduction commands and excluded attempts.


Acentelles:
Consolidated into https://github.com/ICME-Lab/jolt-atlas/tree/review/native-prover (preserved branch; review description in the index), https://github.com/ICME-Lab/jolt-atlas/tree/review/native-verifier (preserved branch; review description in the index).

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `28a3116437032da3d5433f11104f64b1e0c4a4b5` remain available.

PR #346, opened 2026-09-18
Compute independent ZK sumcheck messages in parallel

Acentelles:
## Topic and evidence summary

**Review group:** Prover computation, commitments and scheduling. **Type:** Performance: parallel scheduling.

Compute independent ZK messages in parallel while keeping transcript updates, binding, randomness and shared mutable tables serial.

**Baseline and observed effect.** Isolated opening sumcheck: 1.05861 → 0.3798 s; 0.67881 s less (64.12% reduction; 2.787× baseline/candidate). i7-13700, 16 workers, 4,096 mixed reductions, two samples each. This is the exact upstream fixture; the separate ordinary/deferred-opening experiment is not its baseline.

**Dependencies.** This PR targets #345 (`perf/tiled-native-range-tables`); reviewed head `397338eb0821`, code base `e94b30b2fc1b`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #308, #337, #341, #345, #347, #348, #350, #351, #352, #356, #360.

Native graph proving and opening reduction compute independent ZK sumcheck round messages serially. This change adds an opt-in entry point that computes those messages across workers while preserving their order. Native graph proving and the ZK opening reduction use it. Challenge ingestion, finalization, transcript updates, randomness and opening caches remain serial, including reductions with shared mutable tables.

The existing serial entry point remains available. Exact differential tests compare serialized proof bytes, every transcript update, the following challenge, all BlindFold stage fields and opening claims. Cases cover shared dense and sparse tables, missing entries, mixed domains, the parallel threshold, three worker counts and disabled parallelism.

Validation on Linux for this exact PR source passes 102 native proof and rejection tests, strict library Clippy and formatting. An isolated benchmark of 4,096 mixed opening reductions on an i7-13700 with 16 workers gives these means from serial/parallel/parallel/serial runs.

| Opening sumcheck only | Serial | Parallel messages |
|---|---:|---:|
| Time | 1.05861 s | 0.37980 s |

This is an isolated routine measurement with two samples per variant. Complete model proving remains pending. The PR depends on `perf/tiled-native-range-tables`; it changes no verifier equations or proof encoding.

The base has six unresolved `ACTIVATION_TABLE_BOUND` references in the core workspace ZK tests. That file is unchanged here. This limitation is separate from the passing native library checks above.

PR #345, opened 2026-09-18
Reduce native range table memory with tiled histograms

Acentelles:
## Topic and evidence summary

**Review group:** Prover computation, commitments and scheduling. **Type:** Performance: routine time and memory.

Factor equality weights into tiles and accumulate range digit histograms without a full equality vector.

**Baseline and observed effect.** Isolated range construction: 0.47696 → 0.04425 s; 0.43271 s less (90.72% reduction; 10.78× baseline/candidate). Fixture process peak: 773.36 → 262.84 MiB. i7-13700, 16 workers, 2^24 rows, two samples each. The later application stack includes this change, but has no isolated complete-model attribution for this PR.

**Dependencies.** This PR targets #343 (`perf/native-shared-range-commitments`); reviewed head `e94b30b2fc1b`, code base `28a311643703`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #308, #337, #341, #346, #347, #348, #350, #351, #352, #356, #360.

Native range proving materializes a full equality-weight vector and scans it for each integer digit. This change factors the weights by high and low index bits, accumulates digit histograms within tiles, and combines their exact field sums in parallel. It preserves digit order, the small-input loop, shared-parallelism controls, transcript challenges and verifier equations.

The same native range source is tested on Linux with 101 passing native proof and rejection tests, exact histogram and digit parity across ten range widths and three worker counts, a check with shared parallelism disabled, and strict library Clippy. The parity cases include Boolean evaluation points and mixed domain sizes.

An isolated benchmark over 2^24 rows and 32-bit values on an i7-13700 with 16 workers gives the following means from serial/tiles/tiles/serial runs.

| Range construction only | Original | Tiled |
|---|---:|---:|
| Time | 0.47696 s | 0.04425 s |
| Peak process memory | 773.36 MiB | 262.84 MiB |

The observed routine speedup is 10.78 times. Complete model proving measurements are pending. The native source file matches the measured implementation exactly. This PR depends on `perf/native-shared-range-commitments` and changes one implementation file, including its tests and an ignored benchmark.

Workspace ZK CI currently fails on six unresolved `ACTIVATION_TABLE_BOUND` references in `jolt-atlas-core/src/onnx_proof/e2e_tests.rs`. That file is byte-identical to this PR’s base. PR 344 carries the constant-name correction. The native library checks above pass on this exact PR commit.

PR #337, opened 2026-09-18
Compute one-hot commitments in bounded row segments

Acentelles:
## Topic and evidence summary

**Review group:** Prover computation, commitments and scheduling. **Type:** Performance: kernel time and memory.

Build one-hot commitments in bounded row segments, grouping indices by address within each segment.

**Baseline and observed effect.** One-hot commitment kernel: 3.883 → 2.252 s; 1.631 s less (42.00% reduction; 1.724× baseline/candidate). Process peak including setup decode: 1.145 → 0.716 GiB. 2^22 indices, eight workers, two reversed-order pairs; public commitments are byte-identical. Not a full-model measurement.

**Dependencies.** This PR targets #336 (`feat/native-first-maximum`); reviewed head `fc57b41f19f1`, code base `bc75d22ff010`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #308, #341, #345, #346, #347, #348, #350, #351, #352, #356, #360.

Bound temporary one-hot commitment storage by the row segment width. Each aligned column segment groups its indices by address, runs the existing batched affine row sum, and writes commitments in the original row order. Segments use the existing Rayon pool. The small input path and all commitment equations are unchanged.

Exact tests compare the projective reference across sparse and dense inputs, empty entries, multiple widths and bucket counts, infinity and opposite bases, and one and eight workers. A separate public-setup benchmark produces byte-identical public Dory commitments for both kernels. With 2^22 indices and eight workers, two reversed-order pairs average 3.883 seconds before and 2.252 seconds after; whole-process peak memory, including setup decoding, averages 1.145 GiB and 0.716 GiB. These are kernel measurements and do not establish a full-model speedup.

Depends on #336. The series starts from `perf/round-2`.

## Validation

- Rust 1.97.1, authenticated source export of `fc57b41f19f15a6ea5f02234ec79a316235fda1f`. Workspace crates rebuilt for this branch.
- `cargo nextest` library suites pass 289 tests with `zk` and 164 without it.
- Library Clippy passes with warnings denied.
- Native receipt example Clippy passes with warnings denied.
- The core library compiles with `zk`.

These checks cover the native component API. Complete private ONNX integration, authenticated model IO and representation conversions remain open. Related to #289.

## Repository CI scope

Repository-wide ZK CI remains blocked by the ONNX integration gaps in #289, including existing activation tests that refer to a removed constant and unused softmax helpers. The focused checks above do not establish a complete private ONNX proof.

PR #308, opened 2026-09-16
Build clamp buckets without scanning constant weights

Acentelles:
## Topic and evidence summary

**Review group:** Prover computation, commitments and scheduling. **Type:** Performance.

Determine clamp-bucket membership from operator metadata instead of scanning constant tensors whose numeric bound is discarded.

**Baseline and observed effect.** Avoided scans are structural. The complete round-two integration includes other changes, so it supplies no isolated timing or memory gain for this PR.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `86378ec0977f`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #337, #341, #345, #346, #347, #348, #350, #351, #352, #356, #360.

## Summary

Constructing clamp buckets calls `clamp_value_bound` for every node, then discards the bound and keeps only whether it exists. For constant einsum operands this scans the weight tensors each time the prover or verifier reconstructs the bucket layout.

Determine membership from the operator metadata and use the model's registered `sat_clamp_bits` for the bucket width. Preserve scalar exclusions, zero-scale arithmetic and the handling of incomplete sums. Width derivation, bucket packing, commitments and verification equations are unchanged.

## Validation

- Two bucket membership regressions pass, covering the required operator families, scalar and zero-scale exclusions, and incomplete sums.
- All 13 saturation proof tests pass.
- Matrix multiplication proof tests pass, including positive and negative saturation. The existing non-power-of-two test remains ignored upstream.
- A combined integration with the graph and arithmetic patches passes full Qwen2-0.5B proving, fresh verification, changed-input/output rejection and exact agreement of all 151,936 recorded logits. The integration is not an isolated measurement of this PR; no separate speedup is attributed to this change.
- Formatting, default Clippy/tests, GPT-2, BGE and WASM CI pass.

CI's `zk` job fails on unchanged division and clamp imports in `onnx_proof/zk.rs` and activation constants in `e2e_tests.rs`. This PR changes neither file. The failure log is retained with the Qwen comparison receipts.

PR #347, opened 2026-09-18
Evaluate sparse indicators with tiled equality weights

Acentelles:
## Topic and evidence summary

**Review group:** Prover computation, commitments and scheduling. **Type:** Performance: routine time and memory.

Evaluate sparse indicators using small tiled equality weights rather than an expanded field array.

**Baseline and observed effect.** Isolated sparse evaluation: 0.0763 → 0.011 s; 0.0653 s less (85.58% reduction; 6.936× baseline/candidate). Process peak: 581.61 → 71.26 MiB. i7-13700, 16 workers, 2^24 rows, two samples each; no isolated complete-model gain.

**Dependencies.** This PR targets #345 (`perf/tiled-native-range-tables`); reviewed head `6bc0275457e8`, code base `e94b30b2fc1b`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #308, #337, #341, #345, #346, #348, #350, #351, #352, #356, #360.

Sparse indicator evaluation expands one field element per row before evaluating the resulting dense polynomial. For address domains up to 256 and more than 4,096 rows, this change accumulates row weights by address in small tiles and combines their exact sums using factored equality weights. It avoids the expanded field array and one multiplication per row. Smaller inputs and larger address domains retain the existing path.

The result, transcript, proof format and verifier equations are unchanged. Differential checks cover missing entries, Boolean points, field and optimized challenge types, mixed domains, the threshold, three worker counts and disabled parallelism.

Validation on this exact upstream source passes 168 ordinary library tests with assertions enabled, 102 native proof and rejection tests, strict library Clippy and formatting. The ignored benchmark uses the normal release profile. On an i7-13700 with 16 workers and 2^24 rows, expanded/tiled/tiled/expanded runs give these means.

| Indicator evaluation only | Expanded | Tiled |
|---|---:|---:|
| Time | 0.07630 s | 0.01100 s |
| Peak process memory | 581.61 MiB | 71.26 MiB |

These are isolated routine measurements with two samples per variant, not complete model proving results. A complete model proof is running separately. The PR changes one implementation file, including its differential test and ignored benchmark, and depends on `perf/tiled-native-range-tables`.

The base has six unresolved `ACTIVATION_TABLE_BOUND` references in the core workspace ZK tests. That file is unchanged here. This limitation is separate from the passing native library checks above.


Acentelles:
Consolidated into https://github.com/ICME-Lab/jolt-atlas/tree/review/native-prover (preserved branch; review description in the index).

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `6bc0275457e81e1d55d8df1925b7b547c0c241e4` remain available.

PR #362, opened 2026-09-19
Reclaim working polynomial buffers during opening reduction

Acentelles:
## Topic and evidence summary

**Review group:** Prover memory and ownership. **Type:** Performance: measured complete-prover memory.

Reclaim inactive bound working buffers while preserving the original committed polynomials.

**Baseline and observed effect.** Complete native process peak: 213.8 → 200.97 GiB; 12.83 GiB less (6.00% reduction; 1.064× baseline/candidate). Working capacity after reduction: 92.46 → 4.72 GiB. Whole native time 2,473.98 → 2,473.93 s is essentially unchanged; commitments/proving 1,308.79 → 1,295.82 s. Same Xeon Platinum 8272L, 48 workers, one complete sample each; no general speedup or verifier optimization.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `9834459b7b3f`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #349, #353, #354, #357, #358, #363.

Opening reduction retains each compact working polynomial's original integer coefficients after binding, along with large inactive tails in its field vector. Those allocations remain live while later rounds run.

This change releases the compact coefficients after the first binding and reclaims large field tails when capacity reaches four times the active length. Allocations below one MiB stay in place. It applies to dense openings and the dense working tables of sparse openings. The original committed polynomials remain available for the final commitment proof, and shared openings continue to bind once per round.

Validation on Linux with Rust 1.97.1 and release assertions enabled

- 190 ZK and 160 ordinary library tests pass.
- Independent tests compare every remaining coefficient, full round messages at several evaluation points, shared binding counts and final claims.
- Formatting passes. Library and test Clippy pass with only the two `for_kv_map` diagnostics already present on the exact parent allowed. The original diagnostics are retained.
- All 433 tracked files match the tested snapshot on `perf/round-2` at `cfcffb05dcc76599f4e5b3d01c6e13f5fa653e13`.

The ignored process benchmark retains twelve original compact polynomials while binding their working copies. Its allocator overhead and storage measurements are separate from complete proofs.

A complete native inference comparison now verifies with the original registered parameters and all required consumers and rejection controls, on the same Xeon Platinum 8272L with 48 native workers. One completed sample per variant:

| Measurement | Control | Reclaimed buffers |
|---|---:|---:|
| Commitments and proving | 1308.79 s | 1295.82 s |
[...]

PR #358, opened 2026-09-18
Release dense polynomial inputs after joint construction

Acentelles:
## Topic and evidence summary

**Review group:** Prover memory and ownership. **Type:** Performance: allocation lifetime / owned API.

Consume final joint-opening inputs, move sparse storage and release dense inputs before later allocations, retaining the borrowed API.

**Baseline and observed effect.** No isolated timing or peak result. The subsequent complete policy comparison combines #356, #357 and #358: commitments/proving 29.75 → 30.72 min (+3.25%), whole native command 2,971.14 → 3,065.34 s, peak 213.48 → 213.18 GiB (-0.14%). One sample per configuration on Xeon Platinum 8272L with 48 workers; this establishes neither overall improvement nor a repeatable regression, and cannot isolate any of the three PRs. [Complete comparison](https://github.com/ICME-Lab/zkARc/blob/cc299bc/native-atlas/evidence/2026-09-18-policy-evaluation-combined/results.md).

**Dependencies.** This PR targets #354 (`perf/release-opening-caches`); reviewed head `ad4759e3770e`, code base `cef3f9d6b6dc`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #349, #353, #354, #357, #362, #363.

The final joint opening retains all input polynomials after their dense contributions have been combined. An owned constructor now consumes the map, moves sparse tables into the joint and drops dense inputs before subsequent opening allocations. The native graph uses this path after individual openings finish. Existing borrowed callers retain their API, and both paths use the same Dory opening procedure.

Tests compare both paths with independently evaluated mixed dense and sparse inputs, check ownership without cloning sparse storage, prove and verify real hiding openings, and reject changed commitments and evaluations. No complete time or peak-memory gain is inferred from releasing these inputs.

Exact upstream validation passes formatting, strict ZK library Clippy, 309 ZK library tests and 172 ordinary library tests. All 462 tracked files match the tested Linux tree. This PR is based on #354 and is independent of the evaluation changes.

The inherited workspace ZK CI failure references missing activation test constants and unused softmax methods in jolt-atlas-core. The changed joltworks library has independent passing ZK validation.

PR #357, opened 2026-09-18
Evaluate joint dense values without a full vector copy

Acentelles:
## Topic and evidence summary

**Review group:** Prover memory and ownership. **Type:** Performance: eliminated copy and routine memory.

Evaluate the stored joint dense vector directly through factored equality weights.

**Baseline and observed effect.** Two isolated evaluations of 2^24 field values: 0.36326 → 0.06158 s; 0.30168 s less (83.05% reduction; 5.899× baseline/candidate). Fixture process peak: 1.006 → 0.506 GiB, two samples per variant. The subsequent complete policy comparison combines #356, #357 and #358: commitments/proving 29.75 → 30.72 min (+3.25%), whole native command 2,971.14 → 3,065.34 s, peak 213.48 → 213.18 GiB (-0.14%). One sample per configuration on Xeon Platinum 8272L with 48 workers; this establishes neither overall improvement nor a repeatable regression, and cannot isolate any of the three PRs. [Complete comparison](https://github.com/ICME-Lab/zkARc/blob/cc299bc/native-atlas/evidence/2026-09-18-policy-evaluation-combined/results.md).

**Dependencies.** This PR targets #354 (`perf/release-opening-caches`); reviewed head `4b3d01c7a41b`, code base `cef3f9d6b6dc`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #349, #353, #354, #358, #362, #363.

The joint polynomial evaluator copies its complete dense field vector before evaluating it. This reads the stored vector directly using small factored equality tables. Sparse contributions, variable order and zero padding retain their existing behavior.

Independent component evaluations cover mixed dense and sparse inputs, Boolean and general points, worker counts and disabled parallelism. On the isolated fixture, two evaluations of 2^24 field values average 0.36326 s with the copy and 0.06158 s without it across two samples per variant. Process peak averages 1.006 GiB and 0.506 GiB respectively. These measurements cover the routine and fixture, not a complete prover.

Exact upstream validation passes formatting, strict ZK library Clippy, 308 ZK library tests and 172 ordinary library tests. All 462 tracked files match the tested Linux tree. This PR is based on #354 and is independent of the compact row evaluation change.

The inherited workspace ZK CI failure references missing activation test constants and unused softmax methods in jolt-atlas-core. The changed joltworks library has independent passing ZK validation.

PR #354, opened 2026-09-18
Release finalized opening tables before PCS allocation

Acentelles:
## Topic and evidence summary

**Review group:** Prover memory and ownership. **Type:** Performance: allocation lifetime.

Release finalized opening instances and dense/equality caches before allocating PCS joint tables.

**Baseline and observed effect.** Ownership-release and proof/claim checks pass. No isolated full-model peak-memory or timing improvement is established; combined hint/cache results contain several changes.

**Dependencies.** This PR targets #353 (`perf/shared-opening-hints`); reviewed head `cef3f9d6b6dc`, code base `f8db0be2d5ab`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #349, #353, #357, #358, #362, #363.

Opening finalization extracts scalar claims but leaves dense and equality table owners in its caches. Release the sumcheck instances and both caches before the PCS allocates the joint tables. The extracted claims, transcript and proof format are unchanged.

A ZK opening-reduction test with dense and sparse polynomials checks that the table owners are released and the scalar openings and polynomial order remain intact. Exact Linux validation passes 307 assertion-enabled ZK library tests, 171 ordinary library tests, strict ZK library Clippy and formatting. All 462 tracked source files match the tested tree.

Based on #353. Complete model memory and timing effects are still being measured.

PR #363, opened 2026-09-19
Reduce retained memory for immutable lookup indices

Acentelles:
## Topic and evidence summary

**Review group:** Prover memory and ownership. **Type:** Performance: compact retained indices; public API migration.

Encode immutable index rows with byte/short codes and shared decode tables, keeping direct-storage fallbacks. Vec-specific field users must migrate to views.

**Baseline and observed effect.** Complete native process peak: 200.97 → 163.26 GiB; 37.71 GiB less (18.76% reduction; 1.231× baseline/candidate). Both variants include #362. Commitments/proving 1,295.82 → 1,309.33 s (+1.04%); whole native time 2,473.93 → 2,529.95 s (+2.26%). Same Xeon Platinum 8272L, 48 workers, one sample each. Isolated construction is slower (0.384 → 0.415 s), scanning faster (0.071 → 0.059 s). This is a measured memory saving with time tradeoffs, not a proving-speed claim.

**Dependencies.** This PR targets #288 (`perf/round-2`); reviewed head `593fc22c8fef`, code base `cfcffb05dcc7`. The benchmark baseline is the configuration described above and below, which may include other integrated changes. The topic group does not change merge order. Related PRs: #349, #353, #354, #357, #358, #362.

Lookup indices currently retain four-byte `Option<u16>` entries even for small public tables. Store immutable rows in byte or short codes when the encoded rows and public decode table are smaller, and share that allocation through opening reduction and Dory commitment views. Zero represents absence; the full 65,536-entry address space and small vectors keep direct storage.

The public `OneHotPolynomial::from_indices` constructor still accepts the same vector. Its `nonzero_indices` field now exposes immutable views, so callers using Vec-specific APIs need to migrate to those views. Proof serialization, transcript operations, commitment blinding and verifier predicates are unchanged.

### Validation

- Exact `perf/round-2` base `cfcffb05dcc76599f4e5b3d01c6e13f5fa653e13`, release tests with debug assertions, Rust 1.97.1 on Linux.
- All 191 ZK and 161 ordinary library tests pass. All 434 source files match the validated Linux tree. Formatting and library/test Clippy pass with only the two unchanged parent `for_kv_map` warnings allowed. The strict parent diagnostics match exactly.
- Independent checks cover every address at encoding boundaries, absence, slices, parallel iteration, shared ownership, polynomial evaluations and every binding round. The largest short code and full-size direct fallback are exercised explicitly.
- A separate fixture uses 32 vectors of 2^22 entries for a 256-entry table, with 16 workers. In original, compact, compact, original order, mean peak RSS is 528,608 and 291,060 KiB. Stored arrays occupy 536,870,912 and 268,468,352 bytes, including decode tables. All checksums agree.
- Mean fixture construction is 0.384 and 0.415 seconds; repeated scanning is 0.071 and 0.059 seconds. These short isolated samples do not establish a complete-prover speed or memory improvement.

CI passes formatting, default tests, GPT-2 and BGE proof/verification, and the Wasm build. The workspace ZK job still fails with the same error messages and source locations as the unmodified base. This inherited workspace failure is separate from the passing library checks above.

A complete native inference comparison now verifies with the original registration, both tensor boundaries, required consumer proofs and every rejection control. It uses the same Xeon Platinum 8272L with 48 native workers, one completed sample per variant. Both variants include the separately measured opening-buffer reclamation.

| Complete native invocation | Original indices | Compact indices |
|---|---:|---:|
| Commitments and proving | 1295.82 s | 1309.33 s |
| Whole native command | 2473.93 s | 2529.95 s |
| Peak process memory | 200.97 GiB | 163.26 GiB |
| Unique original index capacity | 69.97 GiB | 33.29 GiB |
| Native proof body | 250,283 bytes | 250,283 bytes |

These single samples establish neither general repeatability nor a broad proving-speed guarantee. Original polynomial capacities remain intact through opening reduction. All proofs and controls pass, complete archives authenticate, and the process and Rust Perfetto traces import without errors or unfinished spans. Vector capacity, allocator accounting and process RSS retain distinct scopes.


Acentelles:
Consolidated into https://github.com/ICME-Lab/jolt-atlas/tree/review/native-prover (preserved branch; review description in the index).

The replacement preserves this change and its recorded evidence. Closing this PR retires the duplicate review queue. The original branch and reviewed head `7455985239ac1f6629a21eb5de51c67aca973dba` remain available.

